PR #6, opened 2019-01-10
Picker enhancement merge

blinkdog:
This PR doesn't properly update the CHANGELOG or bump the version number yet.


dsschult:
My opinion on changelog and version numbers is that during this early development they are meaningless.  It's not like anything really works yet.

lta/log_format.py
```diff
@@ -0,0 +1,36 @@
...
+        super(StructuredFormatter, self).__init__()
+
+    def format(self, record: LogRecord) -> str:
+        # ensure our log message has an ISO 8601 timestamp
+        data = {
+            'timestamp': datetime.utcnow().isoformat()
+        }
+        # copy everything provided to us in the LogRecord object
+        for key in record.__dict__:
+            data[key] = record.__dict__[key]
```
dsschult: Directly using `__` methods is generally a bad idea, as they're supposed to be private.  I think the `vars()` function does what you want here.

PR #10, opened 2019-01-15
basic status handling

dsschult:
This should be good enough to start working on the monitoring, but will obviously get revised over time.

### DIFF
--- a/lta/rest_server.py
+++ b/lta/rest_server.py
@@ -163,6 +163,45 @@ class TransferRequestActionsPopHandler(BaseLTAHandler):
                     break
         self.write({'results': ret})
 
+class StatusHandler(BaseLTAHandler):
+    @lta_auth(roles=['admin', 'user', 'system'])
+    async def get(self) -> None:
+        """Get the overall status of the system"""
+        ret = {}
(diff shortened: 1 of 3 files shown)


lta/rest_server.py
```diff
@@ -163,6 +163,45 @@ class TransferRequestActionsPopHandler(BaseLTAHandler):
                     break
         self.write({'results': ret})
 
+class StatusHandler(BaseLTAHandler):
+    @lta_auth(roles=['admin', 'user', 'system'])
+    async def get(self) -> None:
+        """Get the overall status of the system"""
+        ret = {}
+        health = 'OK'
+        old_data = (datetime.utcnow() - timedelta(seconds=60*5)).isoformat()
```
blinkdog: This seems like another one you'd want a configuration knob for.
Maybe `LTA_STALE_STATUS_SECONDS` ?

dsschult: Yeah, I thought about it, then decided it was not straightforward to implement (challenging to pass that value through to this location) and pushed it down the road.  It'll probably happen in a later refactoring when we impose stricter limits on monitoring quantities.

PR #13, opened 2019-01-18
Refactored heartbeat PATCH to use RestClient

blinkdog:
Modified StructuredFormatter to handle logging Exceptions
Refactored Picker to use RestClient to PATCH status heartbeats
Added coverage tests to test_log_format.py
Refactored test_picker.py to use pytest fixture for picker configuration
Modified loglens to print stack traces supplied by StructuredFormatter
Created lta-db.sh script to run a local instance of LTA REST DB
Added new configuration entities to picker.sh


lta/picker.py
```diff
@@ -16,7 +16,10 @@
 
 EXPECTED_CONFIG = [
     "FILE_CATALOG_REST_URL",
+    "HEARTBEAT_PATCH_RETRIES",
+    "HEARTBEAT_PATCH_TIMEOUT_SECONDS",
```
dsschult: The only comment I have on these two variables is that they might not need to be configurable.  A sensible default is probably good enough.

lta/picker.py
```diff
@@ -53,10 +56,13 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
         self.file_catalog_rest_url = config["FILE_CATALOG_REST_URL"]
-        self.heartbeat_sleep_duration_seconds = int(config["HEARTBEAT_SLEEP_DURATION_SECONDS"])
+        self.heartbeat_patch_retries = int(config["HEARTBEAT_PATCH_RETRIES"])
+        self.heartbeat_patch_timeout_seconds = float(config["HEARTBEAT_PATCH_TIMEOUT_SECONDS"])
+        self.heartbeat_sleep_duration_seconds = float(config["HEARTBEAT_SLEEP_DURATION_SECONDS"])
+        self.lta_rest_token = config["LTA_REST_TOKEN"]
         self.lta_rest_url = config["LTA_REST_URL"]
         self.picker_name = config["PICKER_NAME"]
-        self.work_sleep_duration_seconds = int(config["WORK_SLEEP_DURATION_SECONDS"])
+        self.work_sleep_duration_seconds = float(config["WORK_SLEEP_DURATION_SECONDS"])
```
dsschult: Why float here?  Is this so testing can specify sub-1 second times?

lta/picker.py
```diff
@@ -110,16 +116,18 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
     picker.logger.info(f"PATCH {status_url} - {status_body}")
     try:
-        session = FuturesSession()
-        # r = requests.patch(status_url, data=status_body)
-        r = await asyncio.wrap_future(session.patch(status_url, data=status_body))
-        if (r.status_code < 200) or (r.status_code > 299):
-            picker.logger.error("Unable to PATCH /status/picker with heartbeat")
-            picker.lta_ok = False
-            return picker.lta_ok
+        # TODO: Will probably refactor this into some function -- get a configured RestClient
```
dsschult: You could put it in the picker itself, or pass it in as a parameter to the function.

lta/picker.py
```diff
@@ -110,16 +116,18 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
+        # Use the RestClient to PATCH our heartbeat to the LTA REST DB
+        await rc.request('PATCH', "/status/picker", status_body)
         picker.lta_ok = True
-    except requests.exceptions.ConnectionError:
-        picker.logger.error("ConnectionError trying to PATCH /status/picker with heartbeat")
+    except HTTPError as e:
+        # if there was a problem, yo I'll solve it
+        picker.logger.error("Error trying to PATCH /status/picker with heartbeat")
+        picker.logger.error(f"Error was: '{e}'")
         picker.lta_ok = False
```
dsschult: What if some other exception occurs?  We should still catch those, so we can continue heartbeating.

lta/picker.py
blinkdog: This was motivated first by this idea:

- My configuration had added a true discrete quantity (retries); or, I don't think 2.73 retries has a sensible interpretation. So this one can and should be an `int()`

This made me reconsider the other configuration types:

- Seconds is a continuous quantity where someone could (in theory) provide something like 20.5 seconds. Why they would do it, who knows, but the code would be more tolerant if it also accept float values. So this one can and should be a `float()`.

lta/picker.py
blinkdog: Changed this to `Exception` so we can catch whatever goes wrong.

PR #14, opened 2019-01-28
Implement Picker

blinkdog:
This PR implements the core functionality of the Picker component.

1. Ask the REST DB for one or more TransferRequests to process
2. Ask the File Catalog for all files indicated by the source of the TransferRequest
3. Bulk update the REST DB with new File objects for each file that needs to be transferred
4. Mark the TransferRequest as completed with the REST DB

The coverage tests are not complete yet, but opening the PR allows code review to take place concurrently.

### DIFF
--- a/local-secret
+++ b/local-secret
@@ -1,1 +1,0 @@
-TpnW/sVYy/5gPpTOTSWuv1E4io8HyUa4vja8qZoYqa0=
--- a/local-secret
+++ b/local-secret
@@ -1,0 +1,1 @@
+TpnW/sVYy/5gPpTOTSWuv1E4io8HyUa4vja8qZoYqa0=
--- a/lta/picker.py
+++ b/lta/picker.py
@@ -58,7 +58,7 @@ class Picker:
         logger - The object the picker should use for logging.
         """
         # validate provided configuration
-        for name in EXPECTED_CONFIG.keys():
+        for name in EXPECTED_CONFIG:
             if name not in config:
                 raise ValueError(f"Missing expected configuration parameter: '{name}'")
             if not config[name]:
@@ -84,7 +84,7 @@ class Picker:
         self.last_work_end_timestamp = timestamp
         self.lta_ok = False
         self.logger.info(f"Picker '{self.picker_name}' is configured:")
-        for name in EXPECTED_CONFIG.keys():
+        for name in EXPECTED_CONFIG:
             self.logger.info(f"{name} = {config[name]}")
 
     async def run(self) -> None:
--- a/lta/picker.py
+++ b/lta/picker.py
@@ -9,7 +9,7 @@ import logging
 import platform
 from rest_tools.client import RestClient  # type: ignore
 import sys
-from typing import Dict
+from typing import Any, Dict, List, Union
 from urllib.parse import urljoin
 
 from .config import from_environment
@@ -35,6 +35,9 @@ HEARTBEAT_STATE = [
     "last_work_end_timestamp",
     "lta_ok"
 ]
+
+FileType = Dict[str, Union[str, Dict[Any, Any]]]
+FileList = List[FileType]
 
 
 class Picker:
@@ -154,9 +157,9 @@ class Picker:
         fc_response = await fc_rc.request('GET', f'/api/files?query={query_json}')
         self.logger.info(f'File Catalog returned {len(fc_response["files"])} file(s) to process.')
         # for each file provided by the catalog
-        bulk_create = []
+        bulk_create: FileList = []
         for catalog_file in fc_response["files"]:
-            await self._do_work_catalog_file(lta_rc, tr, fc_rc, dests, catalog_file, bulk_create)
+            bulk_create = bulk_create + await self._do_work_catalog_file(lta_rc, tr, fc_rc, dests, catalog_file)
         # 3. Update the REST DB with Files needed for bundling
         self.logger.info(f'Identified {len(bulk_create)} transfer(s) to add to the REST DB.')
         create_body = {
@@ -175,10 +178,12 @@ class Picker:
         await lta_rc.request('PATCH', f'/TransferRequests/{tr["uuid"]}', complete)
         self.logger.info(f'Done working on TransferRequest {tr["uuid"]}.')
 
-    async def _do_work_catalog_file(self, lta_rc, tr, fc_rc, dests, catalog_file, bulk_create) -> None:
+    async def _do_work_catalog_file(self, lta_rc, tr, fc_rc, dests, catalog_file) -> FileList:
         self.logger.info(f'Processing catalog file: {catalog_file["logical_name"]}')
         # ask the File Catalog for the full record of the file
         fc_response2 = await fc_rc.request('GET', f'/api/files/{catalog_file["uuid"]}')
+        # create a container to hold our results
+        bulk_create: FileList = []
         # for each destination in the transfer request
         for dest in dests:
             # check to see if our full record contains that location
@@ -201,6 +206,8 @@ class Picker:
                 "catalog": fc_response2
             }
             bulk_create.append(file_obj)
+        # return our list to the caller
+        return bulk_create
 
 
 async def patch_status_heartbeat(picker: Picker) -> bool:
--- a/lta/picker.py
+++ b/lta/picker.py
@@ -124,7 +124,7 @@ class Picker:
         response = await lta_rc.request('POST', '/TransferRequests/actions/pop?source=WIPAC', pop_body)
         self.logger.info(f"LTA DB responded with: {response}")
         results = response["results"]
-        if len(results) < 1:
+        if not results:
             self.logger.info(f"No TransferRequests are available to work on. Going on vacation.")
             return
         self.logger.info(f"There are {len(results)} TransferRequest(s) to work on.")
--- a/lta/picker.py
+++ b/lta/picker.py
@@ -3,6 +3,7 @@
 
 import asyncio
 from datetime import datetime
+import json
 from logging import Logger
 import logging
 from rest_tools.client import RestClient  # type: ignore
@@ -14,6 +15,7 @@ from .config import from_environment
 from .log_format import StructuredFormatter
 
 EXPECTED_CONFIG = [
+    "FILE_CATALOG_REST_TOKEN",
     "FILE_CATALOG_REST_URL",
     "HEARTBEAT_PATCH_RETRIES",
     "HEARTBEAT_PATCH_TIMEOUT_SECONDS",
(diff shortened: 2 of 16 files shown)


local-secret
```diff
@@ -0,0 +1 @@
+TpnW/sVYy/5gPpTOTSWuv1E4io8HyUa4vja8qZoYqa0=
```
dsschult: This file should not be included in the repo.

lta/picker.py
```diff
@@ -14,14 +15,17 @@
...
     "HEARTBEAT_PATCH_RETRIES",
     "HEARTBEAT_PATCH_TIMEOUT_SECONDS",
     "HEARTBEAT_SLEEP_DURATION_SECONDS",
     "LTA_REST_TOKEN",
     "LTA_REST_URL",
     "PICKER_NAME",
-    "WORK_SLEEP_DURATION_SECONDS"
+    "WORK_RETRIES",
+    "WORK_SLEEP_DURATION_SECONDS",
+    "WORK_TIMEOUT_SECONDS"
```
dsschult: Can some of these have defaults?  Sane defaults means less cruft when we actually have to configure the service in production.

dsschult: Note that the way I did it so far was to have a second `EXPECTED_CONFIG2 = {}`, or some such.

lta/picker.py
```diff
@@ -79,24 +86,116 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
+        complete = {
+            "complete": {
+                "timestamp": datetime.utcnow().isoformat(),
+                "picker": self.picker_name
+            }
+        }
+        await lta_rc.request('PATCH', f'/TransferRequests/{tr["uuid"]}', complete)
+        self.logger.info(f'Done working on TransferRequest {tr["uuid"]}.')
+
+    async def _do_work_catalog_file(self, lta_rc, tr, fc_rc, dests, catalog_file, bulk_create) -> None:
```
dsschult: It's bad form to have mutating arguments, such as `bulk_create`.  Better to return a list.

lta/rest_server.py
```diff
@@ -87,6 +87,26 @@ def initialize(self, db: Any, check_claims: CheckClaims, *args: Any, **kwargs: A
...
+class FilesActionsBulkCreateHandler(BaseLTAHandler):
+    @lta_auth(roles=['system'])
+    async def post(self) -> None:
+        req = json_decode(self.request.body)
+        if 'files' not in req:
+            raise tornado.web.HTTPError(400, reason="missing files field")
+        if not isinstance(req['files'], list):
+            raise tornado.web.HTTPError(400, reason="files field is not a list")
+        if not req['files']:
+            raise tornado.web.HTTPError(400, reason="files field is empty")
```
dsschult: Recommend consolidating these `if` statements into one.  We shouldn't hand-hold the user at the expense of extra lines of code.  A single reason is sufficient.

lta/rest_server.py
```diff
@@ -87,6 +87,26 @@ def initialize(self, db: Any, check_claims: CheckClaims, *args: Any, **kwargs: A
...
+
+        for xfer_file in req["files"]:
+            xfer_file["uuid"] = uuid1().hex
+            xfer_file["create_timestamp"] = datetime.utcnow().isoformat()
```
dsschult: Use the `now()` function for timestamps, so we stay unified

tests/test_picker.py
```diff
@@ -287,15 +290,19 @@ def test_constructor_state(config):
...
         call('HEARTBEAT_SLEEP_DURATION_SECONDS = 60'),
         call('LTA_REST_TOKEN = fake-lta-rest-token'),
         call('LTA_REST_URL = http://RmMNHdPhHpH2ZxfaFAC9d2jiIbf5pZiHDqy43rFLQiM.com/'),
         call('PICKER_NAME = testing-muh-picker'),
+        call('WORK_RETRIES = 3'),
         call('WORK_SLEEP_DURATION_SECONDS = 60'),
+        call('WORK_TIMEOUT_SECONDS = 30'),
         call('Starting picker work cycle'),
+        call('Asking the LTA DB for a TransferRequest to work on.'),
         call('Ending picker work cycle')
```
dsschult: Personally, I think checking for the config to be specific values is a bad idea.  If we ever change them, this test fails for no reason.  Also, in general I don't really care what the config is as long as the code does what it's supposed to do.

lta/picker.py
blinkdog: Sure. I'm not opposed to sane defaults.

lta/rest_server.py
blinkdog: I have two lines of thinking on this, it depends on if we consider an empty list to be a valid input or not:

1. If an empty list is an error, it's nice to be explicit as to why the request was rejected.

2. If an empty list is not an error, then I'd think to remove Line 98-99. The logic would add 0 files to the database and return 201 with an empty array of UUIDs. The user requested we `bulk_create` nothing, so that's what we do.

lta/rest_server.py
blinkdog: I think we've been using `utcnow()` instead of `now()`:

(env) pmeade@bluebird ~/github/lta $ fgrep -R "datetime.now" lta/*

(env) pmeade@bluebird ~/github/lta $ fgrep -R "datetime.now" tests/*

(env) pmeade@bluebird ~/github/lta $ fgrep -R "datetime.utcnow" lta/*
lta/log_format.py:            'timestamp': datetime.utcnow().isoformat(),
lta/picker.py:        timestamp = datetime.utcnow().isoformat()
lta/picker.py:        self.last_work_begin_timestamp = datetime.utcnow().isoformat()
lta/picker.py:        self.last_work_end_timestamp = datetime.utcnow().isoformat()
lta/picker.py:                "timestamp": datetime.utcnow().isoformat(),
lta/picker.py:            "timestamp": datetime.utcnow().isoformat()
lta/rest_server.py:    return datetime.utcnow().isoformat(timespec='seconds')
lta/rest_server.py:        cutoff_time = datetime.utcnow() - timedelta(hours=self.claim_age)
lta/rest_server.py:            xfer_file["create_timestamp"] = datetime.utcnow().isoformat()
lta/rest_server.py:        old_data = (datetime.utcnow() - timedelta(seconds=60*5)).isoformat()

(env) pmeade@bluebird ~/github/lta $ fgrep -R "datetime.utcnow" tests/*
tests/test_rest_server.py:    request = {'1.1': {'t': datetime.utcnow().isoformat(), 'foo': 'bar'}}
tests/test_rest_server.py:    request2 = {'1.2': {'t': datetime.utcnow().isoformat(), 'baz': 2}}
tests/test_rest_server.py:    request = {'2.1': {'t': (datetime.utcnow() - timedelta(hours=1)).isoformat(), 'foo': 'bar'}}

blinkdog: Ahhh, there is a `now()` function in rest_server. Got it.

lta/rest_server.py
dsschult: I'm inclined to say an empty list is an error, because you can't create nothing (so you shouldn't have called in the first place).  I guess to break the tie, what does @vbrik think?

lta/rest_server.py
dsschult: If you wanted to be explicit, we could rename the function to utcnow.  But since I prefer all times in utc, it just seems like more typing. :smile:

lta/picker.py
```diff
@@ -78,7 +84,7 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
         self.last_work_end_timestamp = timestamp
         self.lta_ok = False
         self.logger.info(f"Picker '{self.picker_name}' is configured:")
-        for name in EXPECTED_CONFIG:
+        for name in EXPECTED_CONFIG.keys():
```
dsschult: Why do you call `.keys()`?  it should work bare.

tests/test_picker.py
blinkdog: This test isn't really written with specific values in mind, but rather that the configuration values are logged as provided when the component starts up. The configuration values that are used in the test are taken from the fixture on Line 43-58, except for one modified in the test itself.

I'm not opposed to having the test use a completely internal configuration (i.e.: don't use the fixture and so avoid being brittle with respect to changes there), but I do think that it is important to test that the component is logging all of its configuration values.

lta/picker.py
blinkdog: Ahhh, I did not realize that. I thought moving from List to Dict meant giving up nice things. Python iterators to the rescue.

tests/test_picker.py
blinkdog: A new test with a better name, a better description, and a locally created configuration object make it a little more clear what the test is doing and why.

lta/rest_server.py
blinkdog: I'm going to leave the error on empty list because it prompts the error case. I'm not sure that it's right to rely on that behavior, but it is what it is for this PR.

vbrik:
A little nitpick: do we have a convention on when to use single and when to use double quotes? Or do they behave differently in python 3?

dsschult:
They behave exactly the same in python.  I usually use single quotes, in case I'm printing another language that requires double quotes inside the string.

The rule: https://www.python.org/dev/peps/pep-0008/#string-quotes

lta/picker.py
```diff
@@ -110,12 +235,11 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
             "timestamp": datetime.utcnow().isoformat()
         }
     }
-    for name in EXPECTED_STATE:
+    for name in HEARTBEAT_STATE:
         status_body[picker.picker_name][name] = getattr(picker, name)  # smh; picker[name]
```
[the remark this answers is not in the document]
dsschult: No, because name is a variable. So it's a dynamic attr request.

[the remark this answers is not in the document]
vbrik: duh :)

blinkdog:
Okay, that should be it. One thing to note is a small change in `_do_work_catalog_file`:

```Python
            for loc in fc_response2["locations"]:
                if (loc["site"] is dest[0]) and (loc["path"].startswith(dest[1])):
                    already_there = True
                    break
```

I changed the "is it already at this destination" check to use `startswith()`, because of the test `test_picker_do_work_catalog_file_fc_yes_result_only_one`.

This test covers the case where the file is requested for archival to NERSC and DESY, and is not present at DESY, but is present at NERSC in a large zip bundle. Note the origin directory (/data/exp...) does not match the destination directory (/tmp/this/is...).

I'm not 100% sure if `startswith()` is the right way to handle this, but I think it's good enough for this PR. If we need to change it, I think we should open an issue an pin down exactly what our policies are w/rt archival paths and bundling.

PR #23, opened 2019-02-06
Implements /Files/actions/pop for the LTA DB

blinkdog:
Added some documentation in doc/admin.md about sites schema
Added etc/sites.json as a default site configuration for LTA DB
Organized imports in rest_server.py
Organized configuration keys in rest_server.py
Added new configuration key "LTA_SITE_CONFIG" to rest_server.py
Added helper function site() to rest_server.py
Added argument 'sites' to BaseLTAHandler
Created FilesActionsPopHandler to implement pop functionality
Updated some requirements.txt dependencies to latest versions
Added dependency 'binpacking' to setupenv.sh and requirements.txt
Added ephemeral port fixture to test_monitoring.py
Bumped up timeout in test_rest_server.py
Added tests for /Files/actions/pop functionality
Added show-how-it-works tests for binpacking dependency


lta/rest_server.py
```diff
@@ -218,6 +228,99 @@ class FilesHandler(BaseLTAHandler):
...
+            if not (site(db_file["source"]) == src):
+                continue
+            if not (site(db_file["dest"]) == dest):
+                continue
+            if not (site(db_file["status"]) == "waiting"):
+                continue
+            src_tuples.append((uuid, db_file["catalog"]["file_size"]))
+        # we can't bundle nothing
+        if len(src_tuples) < 1:
+            self.set_status(200)
```
dsschult: `set_status(200)` is not necessary, as that is the default status.  only necessary when you want a different number.

tests/test_binpacking.py
```diff
@@ -0,0 +1,52 @@
+# test_binpacking.py
```
dsschult: Why are we testing `binpacking`?  This is an external import, not our code.

blinkdog: Just some tests to see/show how the library works.

dsschult: While it's interesting, it doesn't really belong in our repo.  If it's really useful, it should go upstream.

lta/rest_server.py
blinkdog: I added a commit to remove the unnecessary calls.

tests/test_binpacking.py
blinkdog: Pushed another commit to remove that file.

blinkdog:
I pushed a commit to reorganize the imports of `rest_server.py`.
I think that's the way they should go according to PEP8?

PR #24, opened 2019-02-20
LTA DB uses MongoDB

blinkdog:
Refactored rest_server.py to use MongoDB
Modified unit tests to reset the state of MongoDB as necessary for testing
Updated requirements to latest versions
Removed package `binpacking` from requirements

Bundling prefers locality to minimizing number of bundles
LTA DB requires a functional MongoDB as the backing store
Warning: Testing will wipe your MongoDB collections; don't run tests against production!


lta/rest_server.py
```diff
@@ -269,79 +278,108 @@ class FilesActionsPopHandler(BaseLTAHandler):
...
+        for build_file in build_files:
+            build_file["claimant"] = pop_body
+            build_file["claimed"] = True
+            build_file["claim_time"] = claim_time
+            build_file["status"] = "processing"
+            update_query = {"uuid": build_file["uuid"]}
+            update_doc = {"$set": build_file}
+            ret = await sdf.find_one_and_update(update_query, update_doc)
+            if not ret:
+                raise tornado.web.HTTPError(500, reason=f"unable to claim file for bundling")
```
dsschult: This shouldn't be a fatal error.  We need to allow two pop operations to happen in parallel, and they might try to step on each other.

lta/rest_server.py
```diff
@@ -392,28 +437,46 @@ class TransferRequestSingleHandler(BaseLTAHandler):
...
-        self.db['TransferRequests'][request_id].update(req)
-        self.write({})
+        # self.db['TransferRequests'][request_id].update(req)
+        # self.write({})
+        sbtr = self.db.TransferRequests
+        query = {"uuid": request_id}
+        update = {"$set": req}
+        ret = await sbtr.find_one_and_update(query,
+                                             update,
+                                             REMOVE_ID,
```
dsschult: Be explicit about what parameter `REMOVE_ID` is.
```suggestion
                                             projection=REMOVE_ID,
```

lta/rest_server.py
```diff
@@ -427,20 +490,35 @@ class TransferRequestActionsPopHandler(BaseLTAHandler):
...
+            row["claimed"] = True
+            row["claim_time"] = now()
+            update_query = {"uuid": row["uuid"]}
+            update_doc = {"$set": row}
+            ret2 = await sdtr.find_one_and_update(update_query,
+                                                  update_doc,
+                                                  REMOVE_ID,
+                                                  return_document=AFTER)
+            if not ret2:
+                raise tornado.web.HTTPError(500, reason="unable to update transfer request")
```
dsschult: Similar to above pop function, needs to work properly in parallel instead of raising an error.

blinkdog:
Yes, the tests require a running MongoDB.

The tests can be mocked, but as they are now they rely on the implicit state of the system. Mocking will make a lot of that state explicit, so it's maybe 4-8 hours of work and doubling the size of the tests. Not necessarily a bad thing, just a question if we want to invest the effort.

dsschult:
It's not a bad thing.  I was just wondering why it said "all checks have passed."  Drilling down, the snake rebuild is failing, but still returns 0.  So something's wrong there.

blinkdog:
Okay, something of an overhaul.

Minor stuff
- Some commented out memory-DB lines were removed
- Some logging was added where things were created/updated/deleted
- Some functions that were no longer called were removed

Major stuff
- POST /Files/actions/pop was rewritten for Mongo transaction semantics
- POST /TransferRequests/actions/pop was rewritten for Mongo transaction semantics
- snake script was updated to bring >0 return codes to the top level

Coverage for rest_server.py isn't 100%. The only major gap I see is a test for when Mongo has a file that is too big to go into a bundle. I'm ambivalent on writing it, so I thought I'd wait for code review feedback before committing to it.

blinkdog:
Oh! And I reduced the default limit for POST /TransferRequests/actions/pop from 10 to 1.
I think it would be better for a Picker to take one at a time by default.
A Picker can always take a big chunk of 10 by adding `&limit=10` if it wants to.

lta/rest_server.py
```diff
@@ -269,79 +276,130 @@ class FilesActionsPopHandler(BaseLTAHandler):
...
+            "status": "waiting"
+        }
+        sort = [('catalog.logical_name', ASCENDING)]
+        bundle_files = []
+        bundle_full = False
+        bundle_size = 0
+        async for row in sdf.find(filter=query,
+                                  projection=REMOVE_ID,
+                                  skip=skip,
+                                  limit=limit,
```
dsschult: Don't limit the query, instead limit on the actual length of the bundle (maybe in line 324).

lta/rest_server.py
```diff
@@ -392,55 +453,86 @@ class TransferRequestSingleHandler(BaseLTAHandler):
...
-                ret.append(req)
-                req['claimant'] = pop_body
-                req['claimed'] = True
-                req['claim_time'] = now()
-                limit -= 1
-                if limit <= 0:
-                    break
+        async for row in sdtr.find(filter=query,
+                                   projection=REMOVE_ID,
+                                   limit=limit):
```
dsschult: Don't limit the query, instead limit on the actual length of the `ret` array.

dsschult:
After those two small fixes, then just need to fix the test suite so it passes.

blinkdog:
Okay, now to mock out the Mongo dependency ...

dsschult:
Suggestion: just modify the circleci config so it uses mongodb.  Example from file catalog:
```
        docker:
            - image: circleci/python:3.7.2
              environment:
                TEST_DATABASE_URL: localhost:27017
            - image: circleci/mongo:3.7.9-ram
```


Also, fix the setupenv.sh script by using:
`pip install -r requirements.txt`

setupenv.sh
```diff
@@ -4,5 +4,5 @@ virtualenv -p python3 env
 echo "unset PYTHONPATH" >> env/bin/activate
 . env/bin/activate
 pip install --upgrade pip
-pip install binpacking flake8 flake8-docstrings mypy prometheus_client pytest pytest-asyncio pytest-cov pytest-mock requests tornado wheel
 pip install git+https://github.com/WIPACrepo/rest-tools
```
dsschult: now that you're installing from requirements.txt, can you drop this line?

blinkdog: I think so, as long as rest-tools gets installed first. Otherwise, pip will complain that rest-tools can't be found in PyPI

dsschult: Right, I fixed that by installing it in pip with the '-e' option:
```
pip install -e git+https://github.com/WIPACrepo/rest-tools#egg=rest-tools
```
so when you call `pip freeze` it actually appears correctly.

blinkdog: Ahhh, okay, now that I do that, I get that same change in `requirements.txt`:
```
requests-mock==1.5.2
requests-toolbelt==0.9.1
-e git+https://github.com/WIPACrepo/rest-tools@66016820552c6e3452a2d45413772ca53fc4c1e2#egg=rest_tools
setuptools==40.8.0
six==1.12.0
```
I'll make these changes on the /Bundles PR.

PR #25, opened 2019-02-25
Bundle route support merge

blinkdog:
From the [REST API spec](https://docs.google.com/document/d/1KBAvnAQ53rhtts1nSC224_uOyKsC-RqgKWuCTLhq4vo/edit):

GET        /Bundles?query            Get list of Bundles
POST        /Bundles                Create a Bundle

POST        /Bundles/actions/bulk_create        Add bundles to the master transfer list
POST        /Bundles/actions/bulk_delete        Remove bundle from master transfer list
POST        /Bundles/actions/bulk_update        Update bundle in the master transfer list
POST        /Bundles/actions/pop?query        Consume bundle for transfer

DELETE    /Bundles/{uuid}            Delete a Bundle
GET        /Bundles/{uuid}            Get Bundle data
PATCH    /Bundles/{uuid}            Update the state of a Bundle

Implements these routes, but no unit tests yet. I'm headed to the Rucio conference this week, so better to get my code in before the plane crashes. :-)


blinkdog:
Also fixes a small bug in TransferRequests pop; before it could not reclaim old/expired claims.
I adjusted the query to allow this reclaiming.
Sorry, no unit test for that one either.

dsschult:
Can you clarify what the manifest record is?  I want to make sure I understand what that field does.

blinkdog:
Added str2bool package dependency
bulk_create now sets claimed = False by default for bundles
GET /Bundles translates verified into proper boolean values for query
POST /Bundles/actions/pop now considers status when popping
POST /Bundles/actions/pop prefers to pop the oldest bundle first
Updated to latest versions of packages: attrs, numpy, pyflakes, tornado
Added a bunch of tests for /Bundles routes

lta/rest_server.py
```diff
@@ -17,7 +17,8 @@
 from pymongo import MongoClient
 from rest_tools.client import json_decode  # type: ignore
 from rest_tools.server import authenticated, catch_error, RestHandler, RestHandlerSetup, RestServer  # type: ignore
-import tornado.web  # type: ignore
+from str2bool import str2bool  # type: ignore
```
dsschult: I object to a dependency this simple, just on principle.

lta/rest_server.py
```diff
@@ -286,15 +288,18 @@ class BundlesActionsPopHandler(BaseLTAHandler):
...
                     {"claim_time": {"$lt": f"{old_age}"}}
                 ]}
             ]
         }
+        sort = [('create_timestamp', ASCENDING)]
         ret = []
         async for row in sdb.find(filter=query,
-                                  projection=REMOVE_ID):
+                                  projection=REMOVE_ID,
+                                  sort=sort):
```
dsschult: Since I just went through this for IceProd, make sure there's an index for that sort.

tests/test_rest_server.py
```diff
@@ -654,3 +654,431 @@ def client(role='admin', timeout=0.1):
     for res_file in results:
         total_size += res_file["catalog"]["file_size"]
     assert total_size < 1000000000
+
+@pytest.mark.asyncio
+async def test_bundles_bulk_crud(mongo, rest):
+    """Check CRUD semantics for bundles."""
+    mongo.Bundles.delete_many(ALL_DOCUMENTS)
```
dsschult: is there a better way to make this automated?  maybe put this in the fixture, so it's not in every test?

blinkdog:
>Can you clarify what the manifest record is? I want to make sure I understand what that field does.

And just to document what I said in the Dev Sprint meeting this week:
- JADE LTA bundles include a `{UUID}.metadata.json` in the bundle.
- `{UUID}.metadata.json` has the following schema:

```JSON
{
    "component": "bundler",
    "version": 1,
    "referenceUuid": "81e832d8-0a99-4d86-972c-e27ea8608906",
    "dateCreated": "2019-03-05T13:43:21Z",
    "destDir": "/data/exp/IceCube/blah/blah",
    "files": [
        {
            "fileName": "blah_blah.txt",
            "fileSize": 123456,
            "srcDir": "/data/exp/IceCube/blah/blah",
            "uuid": "997bf092-cfe7-4afb-ab9d-ef6f1f06fa5b",
            "checksum": "b3dc6ac98bcac8ce320a7fb836531b683829efcb1b491b8ad9ecb712ee2f6b121b5d42da217f6d95256764e7eb4ebf2662131a373b14c64e81cddb73f2d94072",
            "checksumType": "sha512",
            "DIF_Plus": {
                // JSON representation of the XML from .meta.xml
            }
        }
    ]
}
```

We don't have to use this exact format, but I think something similar should be created and stored with the bundle record in Mongo. The only part that might be **really** annoying is the `DIF_Plus` field. In Java, I wrote some type converter plugins for Google's GSON library:

https://github.com/WIPACrepo/kanoite/tree/master/src/main/java/edu/wisc/icecube/jade/gson

My guess is Python has gotta have an XML -> JSON conversion package somewhere in PyPI. If it'll handle our schema nicely or not is an open question.

I'd actually consider making a tiny Java microservice with a single `POST /convert` route taking an XML body and returning JSON.
Cons: It's an external dependency for LTA
Pros: It'd take 2-4 hours to create and a stateless microservice usually doesn't require much care and feeding once it's wrapped up in a container.

Anyway, this manifest schema question will need to be answered in this work:
https://wipacit.atlassian.net/browse/DEV-283

dsschult:
@blinkdog you mean this package? https://pypi.org/project/xmljson/

blinkdog:
Ahhh yes, `xmljson` looks like what we need.
The real question is how well it will handle our ugly schema:
https://github.com/WIPACrepo/kanoite/blob/master/src/main/xsd/IceCubeDIFPlus.xsd

lta/rest_server.py
blinkdog: Added for both Files and Bundles:
https://github.com/WIPACrepo/lta/pull/25/commits/fa01a201a67d3eb14cfd1043c8f54409b58007e2

lta/rest_server.py
blinkdog: But have we truly lived until we experience LTA stop working when a developer unpublishes the left_pad package? :-)

tests/test_rest_server.py
blinkdog: I'm unclear as to what we want to refactor here?
The data sets used by testing functions?
Or make some omni-CRUD routine where we provide the type?

dsschult: the line `mongo.Bundles.delete_many(ALL_DOCUMENTS)`, which gets repeated at the start of every test.

dsschult: would be nice to assume that the DB is empty at the start of a test (and for the mongo fixture to "make it so")

blinkdog: Ahhh, OK. That should be easy enough.

blinkdog: Done: https://github.com/WIPACrepo/lta/pull/25/commits/76fc3e039b8ace59f905d1b1e328cd1cccc6129f

lta/rest_server.py
blinkdog: Done: https://github.com/WIPACrepo/lta/pull/25/commits/76fc3e039b8ace59f905d1b1e328cd1cccc6129f

PR #27, opened 2019-03-24
Implement Bundler component

blinkdog:
There are lots of edits here, so I've written an annotated guide to the changes:

modified: README.md
- Expanded documentation on creating a local testing environment

new file:   bundler.sh
- Created support script to start a Bundler component from the command line

modified:   doc/admin.md
- Modified REST DB references to LTA DB

modified:   etc/site.json
- Modified the bundle_size value of NERSC for testing purposes

modified:   lta-db.sh
- Modified the support script to make all configurables explicit

new file:   lta/bundler.py
- Implemented the Bundle component

new file:   lta/lta_types.py
- Refactored the File Catalog and LTA DB types to their own Python module

modified:   lta/picker.py
- Rearranged imports in standard format
- Modified REST DB references to LTA DB

modified:   lta/rest_server.py
- Modified REST DB references to LTA DB
- Simplified str2bool implementation per discussion with David

modified:   picker.sh
- Modified default values in support script for testing purposes

modified:   requirements.txt
- Updated to latest version of several Python packages

modified:   resources/make_token.py
- Updated the token printing due to an update in rest_tools

modified:   resources/make_transfer_request.py
- Modified REST DB references to LTA DB

new file:   resources/test_data_helper.py
- Created a support Python program to help with adding and removing testing data

new file:   test-data-helper.sh
- Created a support script to help with adding and removing testing data

new file:   test-data-reset.sh
- Created a support script to help with priming test data

new file:   tests/test_bundler.py
- Adds unit tests for the Bundler component

modified:   tests/test_picker.py
- Modified REST DB references to LTA DB

modified:   tests/test_rest_server.py
- Modified REST DB references to LTA DB

The last unit test for the Bundler is kinda bogus. It does allow coverage, but it doesn't really test anything very well. I can go back and rewrite this if necessary, but I struggled for awhile trying to find a good construction for mocking out the `with ZipFile`; what an incredible pain to test `with` clauses.

Other than the Bundler component itself, I've added some support scripts for testing data, and I've updated the instructions in `README.md` so that setting up a local test environment is easier. I ran plenty of full integration tests, from adding files to the catalog, all the way to finished ZIP bundle. So I'm confident that the system works as intended.

One of the odd points was that ZIP will keep the full path of what is written to the zip file, unless you give it a second argument. I checked the final result with `unzip -l 1d2d78bd-3aba-4af6-adc6-43ad42e1c882.zip` to make sure that simple names and no paths were added to the final bundle.


lta/bundler.py
```diff
@@ -0,0 +1,324 @@
...
+                }
+            },
+            "files": [x["catalog"]["uuid"] for x in results]
+        }
+        await lta_rc.request('POST', '/Files/actions/bulk_update', update_body)
+        # 4.2. Remove the files from the DB
+        delete_body = {
+            "files": [x["catalog"]["uuid"] for x in results]
+        }
+        await lta_rc.request('POST', f'/Files/actions/bulk_delete', delete_body)
```
dsschult: Why are we deleting the files here?  What if the files go to two different destinations - won't this cause the second destination to miss the files?

I thought the `deleter` module was supposed to delete files, after they were verified to be at the destinations.

lta/bundler.py
```diff
@@ -0,0 +1,324 @@
...
+def sha512sum(filename: str) -> str:
+    """Compute the SHA512 hash of the supplied filename."""
+    h = hashlib.sha512()
+    b = bytearray(128*1024)
+    mv = memoryview(b)
+    with open(filename, 'rb', buffering=0) as f:
+        # Known issue with MyPy: https://github.com/python/typeshed/issues/2166
+        for n in iter(lambda: f.readinto(mv), 0):  # type: ignore
+            h.update(mv[:n])
+    return h.hexdigest()
```
dsschult: `sha512sum` should probably be factored out into a util file, since we'll need it in a few other places.

lta/rest_server.py
```diff
@@ -41,26 +41,19 @@
...
+def boolify(value: str) -> bool:
+    """Convert a string into a True or False value."""
     if isinstance(value, str):
         value = value.lower()
-        if value in STR2BOOL_TRUE_SET:
+        if value in TRUE_SET:
             return True
-        if value in STR2BOOL_FALSE_SET:
-            return False
+    return False
```
dsschult: A shorter implementation of `boolify` might be:

```python
return isinstance(value, str) and value.lower() in TRUE_SET
```

tests/test_bundler.py
```diff
@@ -0,0 +1,445 @@
...
+
+    This is useful for creating object literals to be used as return
+    values from mocked API calls.
+
+    Source: https://stackoverflow.com/a/3335732
+    """
+
+    def __init__(self, **kwds):
+        """Add attributes to ourself with the provided named arguments."""
+        self.__dict__.update(kwds)
```
dsschult: These two helper classes have appeared a few times, which means they probably need to be in a util file for testing.

dsschult:
General note: can we have smaller pull requests?  It took me an hour to review this, and I skipped over most of the test details (I assume since it passed it's OK).

lta/bundler.py
blinkdog: Got rid of this code block, so no more deleting.

lta/bundler.py
blinkdog: Refactored out into `lta/crypto.py` and added a unit test for the function.

lta/rest_server.py
blinkdog: Used this implementation verbatim.

tests/test_bundler.py
blinkdog: These got refactored out to `tests/test_util.py` and the unit tests for Picker, Bundler, and monitoring were refactored to use them.

blinkdog:
I also named the internal functions of the Bundler component more explicitly.
I left `_do_work` because I would like to refactor out a common `LongTermArchiveComponent` class at some point in the future, and `_do_work` is my abstract name for performing a single work cycle, whatever that happens to be for the component.

PR #28, opened 2019-03-24
Implemented ltacmd request command

blinkdog:
Implements the following:

`ltacmd display-config`

`ltacmd request ls`
`ltacmd request new`
`ltacmd request status`

However, `ltacmd request new` with the parameter `--fc-query` might be a stretch too far for reasons I've noted in the Google Doc:

https://docs.google.com/document/d/1guHTvpeC1RfmVGwa-8LjDFpWVngiWFyTmeo1dR4rvPU/edit


resources/lta_cmd.py
```diff
@@ -0,0 +1,205 @@
...
+                               nargs='?')
+    parser_status.add_argument("--json",
+                               help="display output in JSON",
+                               action="store_true")
+    parser_status.set_defaults(func=status)
+
+    # parse the provided command line arguments and call the function
+    args = parser.parse_args()
+    exit_code = await args.func(args)
+    sys.exit(exit_code)
```
dsschult: This is considered not-pythonic.  The "correct" way is to just raise an exception and let that return a non-zero exit code.

resources/lta_cmd.py
```diff
@@ -0,0 +1,205 @@
...
+                print(f"{short_id}\t{create_time}\t{path}\t{source} -> {dests}")
+    except Exception as e:
+        print(e)
+        return -1
+    return 0
+
+
+async def request_new(args) -> int:
+    # configure a RestClient from the environment
+    rc = RestClient(args.config["LTA_REST_URL"], token=args.config["LTA_REST_TOKEN"])
```
dsschult: The RestClient configuration smells like duplicate code, and should be factored out somewhere.  Some options:
* an external setup function
* making a class for all the functions, and doing "setup" stuff in the `__init__`
or something else if you can think of anything better.

resources/lta_cmd.py
```diff
@@ -0,0 +1,205 @@
...
+    if args.json:
+        print_dict_as_pretty_json(args.config)
+    else:
+        for key in args.config:
+            print(f"{key}:\t\t{args.config[key]}")
+    return 0
+
+
+def print_dict_as_pretty_json(d: Dict) -> None:
+    print(json.dumps(d, indent=4, sort_keys=True))
```
dsschult: I'm unsure we need json output from the lta cmd.  What might be more useful is to make each of these functions callable (as part of a public python api).

Also, having simple functions as examples of interacting with the API show others how to do the same (so if someone wanted to write this in a different language, it should be easy).

resources/lta_cmd.py
```diff
@@ -0,0 +1,205 @@
...
+    return 0
+
+
+async def request_status(args) -> int:
+    rc = RestClient(args.config["LTA_REST_URL"], token=args.config["LTA_REST_TOKEN"])
+    try:
+        response = await rc.request("GET", "/TransferRequests")
+        results = response["results"]
+        for x in results:
+            if x["uuid"].startswith(args.uuid):
```
dsschult: This makes me wonder if we should make the uuid as an optional parameter to `/TransferRequests`.

resources/lta_cmd.py
blinkdog: I've removed the exit coding handling, and the try-catch blocks.
Let exceptions raise where they may...

resources/lta_cmd.py
blinkdog: Refactored this so that RestClient is configured in main and passed along via `set_defaults`.

resources/lta_cmd.py
blinkdog: I don't mind adding a Python API ala some `LongTermArchiveCommand` class or something, but I think this should probably be a separate card in JIRA.

As for the JSON output, I think it's useful for robots or other scripts coming at LTA from the command line. I can strip it out if we're really opposed to it, but my preference is to keep it.

resources/lta_cmd.py
blinkdog: That would be the right way to implement this. I've opened an issue for it:
https://github.com/WIPACrepo/lta/issues/29

Unfortunately, even with it, I don't think it gets us away from the loop. The short ID (depending on how short it is, for example `"2"`) may return multiple objects, so we're still bound to iterating over the potential results.

PR #30, opened 2019-04-01
Implement LTA lifecycle commands

blinkdog:
By the magic of semaphores, we can now start and stop components with `ltacmd`.

Commands are:
`ltacmd drain` - Drain the currently running component, then stop
`ltacmd start <component>` - Start the named component
`ltacmd stop` - Stop the named component

Bugs are:
`ltacmd kill` - We save the PID generated by `ltacmd start`, but `kill -9` only gets the bash script, not the Python process. Still researching a way to make this work nicely.


lta/bundler.py
```diff
@@ -233,6 +235,21 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
         }
         await lta_rc.request('POST', '/Files/actions/bulk_update', update_body)
 
+
+def check_drain_semaphore() -> bool:
+    """Check if a drain semaphore exists in the current working directory."""
+    cwd = os.getcwd()
+    drain_filename = os.path.join(cwd, DRAIN_SEMAPHORE_FILENAME)
+    return Path(drain_filename).exists()
```
dsschult: This function seems like duplicate code, since it's in every module.

lta/bundler.py
```diff
@@ -233,6 +235,21 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
+
+
+def check_stop_semaphore() -> bool:
+    """Check if a stop semaphore exists in the current working directory."""
+    cwd = os.getcwd()
+    drain_filename = os.path.join(cwd, STOP_SEMAPHORE_FILENAME)
+    return Path(drain_filename).exists()
```
dsschult: As I point out later, this function seems unnecessary.

lta/bundler.py
```diff
@@ -265,10 +282,25 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
+async def lifecycle_loop(bundler: Bundler) -> None:
+    """Run a check for a stop semaphore as an infinite loop."""
+    bundler.logger.info("Starting lifecycle loop")
+    while True:
+        # if there is a stop semaphore, terminate the program
+        if check_stop_semaphore():
+            bundler.logger.info("Component stopped; shutting down.")
+            sys.exit(0)
+        # sleep until we check again
+        await asyncio.sleep(1.0)
```
dsschult: Might be easier to just get a SIGTERM instead of checking explicitly for something every second.

lta/bundler.py
```diff
@@ -265,10 +282,25 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
...
+
+
 async def status_loop(bundler: Bundler) -> None:
     """Run status heartbeat updates as an infinite loop."""
     bundler.logger.info("Starting status loop")
     while True:
+        # if there is a drain semaphore, stop sending status updates
+        if check_drain_semaphore():
+            break
```
dsschult: Could you just do:

```python
while not check_drain_semaphore():
```

lta/bundler.py
```diff
@@ -279,6 +311,10 @@ def __init__(self, config: Dict[str, str], logger: Logger) -> None:
     """Run bundler work cycles as an infinite loop."""
     bundler.logger.info("Starting work loop")
     while True:
+        # if there is a drain semaphore, don't do any additional work
+        if check_drain_semaphore():
+            bundler.logger.info("Component drained; shutting down.")
+            sys.exit(0)
```
dsschult: Same as above, putting it in the `while` condition.

lta/lta_cmd.py
```diff
@@ -191,17 +288,32 @@ def date_ok(d: str) -> bool:
...
                                help="display output in JSON",
                                action="store_true")
     parser_status.set_defaults(func=status)
 
+    # define a subparser for the 'stop' subcommand
+    parser_stop = subparser.add_parser('stop', help='stop the component')
+    parser_stop.add_argument("component",
+                             choices=COMPONENT_NAMES,
+                             help="optional LTA component",
+                             nargs='?')
```
dsschult: If we're just going to ignore the component name, why take it as an argument?

lta/bundler.py
blinkdog: This is one that will probably end up refactored in DEV-605:
https://wipacit.atlassian.net/browse/DEV-605

lta/bundler.py
blinkdog: Yep, this one went away.

lta/bundler.py
blinkdog: Yep, this too also went away.

lta/bundler.py
blinkdog: I liked this better, so I did end up using it verbatim.

lta/bundler.py
blinkdog: Yep.

lta/lta_cmd.py
```diff
@@ -31,6 +33,33 @@
 }
 
 
+def deamon_for(component: str) -> Daemon:
+    """Create a Daemon object for the specified component name."""
+    pidfile = pid_filename(component)
+    runner = __import__(f"lta.{component}", fromlist=['']).runner
```
dsschult: Should use `importlib.import_module`.

lta/lta_cmd.py
blinkdog: I ended up making this component argument non-optional.
On the upside, multiple components can be started from the same repo directory.

lta/lta_cmd.py
blinkdog: Fixed this one up.

PR #32, opened 2019-04-08
Replicator Component

blinkdog:
This one isn't done, I'm just opening the PR to provide visibility.
Visibility allows review of smaller chunks of progress for faster feedback.

Implements the shell of the Replicator component for LTA.
The core work (talking to the LTA DB, talking to Rucio) remains to be filled in.


lta/replicator.py
```diff
@@ -0,0 +1,117 @@
...
+    It registers the bundles with Rucio. It updates the Bundle and the
+    corresponding TransferRequest in the LTA DB with a 'transferring' status.
+    """
+
+    def __init__(self, config: Dict[str, str], logger: Logger) -> None:
+        """
+        Create a Replicator component.
+
+        config - A dictionary of required configuration values.
+        logger - The object the bundler should use for logging.
```
blinkdog: Comment refers to bundler, should refer to replicator.

blinkdog:
JIRA Crosslink: https://wipacit.atlassian.net/browse/DEV-582

lta/replicator.py
```diff
@@ -0,0 +1,117 @@
...
+        # TODO: Perform the work against rucio
+        # 1. Pop bundles from the LTA DB
+        # 2. for each bundle
+        # 2.1. upload and regsiter the bundle to rucio
+        #     rucio upload --rse $RSE --scope SCOPE --register-after-upload --pfn PFN --name NAME /PATH/TO/BUNDLE
+        # 2.2. for each destination site
+        # 2.2.1. add the BUNDLE_DID from 2.1 to the replica
+        #     rucio attach DEST_CONTAINER_DID BUNDLE_DID
+        # 2.3. update the Bundle in the LTA DB; registration information
+        # 2.4. update the TransferRequest in the LTA DB; state: [None] -> [Transferring]
```
dsschult: Use Rucio REST API.

blinkdog: The Rucio client is a pretty thick wrapper over the REST API.

I think this is the entry point for the `rucio upload` command:
https://github.com/rucio/rucio/blob/master/bin/rucio#L828

Which goes down through several layers:
https://github.com/rucio/rucio/blob/master/lib/rucio/client/uploadclient.py#L71
https://github.com/rucio/rucio/blob/master/lib/rucio/client/uploadclient.py#L204

https://github.com/rucio/rucio/blob/master/lib/rucio/rse/rsemanager.py#L350
https://github.com/rucio/rucio/blob/master/lib/rucio/rse/rsemanager.py#L429
https://github.com/rucio/rucio/blob/master/lib/rucio/rse/rsemanager.py#L471

https://github.com/rucio/rucio/blob/master/lib/rucio/rse/protocols/protocol.py#L376
https://github.com/rucio/rucio/blob/master/lib/rucio/rse/protocols/sftp.py#L105

I'm not sure using the REST API is practical for our use case.

dsschult: I don't think tracing the cli code will lead to proper REST API solutions.  Might be better to look at the REST API docs directly:
https://rucio.readthedocs.io/en/latest/rest.html

blinkdog: https://rucio.readthedocs.io/en/latest/restapi/did.html#post--dids-(scope)-(name)

blinkdog: *sigh* ...

https://github.com/rucio/rucio/blob/master/lib/rucio/core/did.py#L162

And immediately the scene from Kill Bill comes to mind, where Lucy Liu says "You didn't think it was going to be that easy, did you?"

blinkdog: Well, going back to the upload case and tracing through the code, it appears this method is responsible for getting a File into Rucio:

https://github.com/rucio/rucio/blob/master/lib/rucio/client/replicaclient.py#L182

This appears to be a `POST /replicas` with the following data:
` data = {'rse': rse, 'files': files, 'ignore_availability': ignore_availability}`

Where files is documented like this:
```
:param files: The list of files. This is a list of DIDs like :
[{'scope': <scope1>, 'name': <name1>}, {'scope': <scope2>, 'name': <name2>}, ...]
```

The dictionary it expects is documented in the `add_replica` method above:
```
        dict = {'scope': scope, 'name': name, 'bytes': bytes, 'meta': meta, 'adler32': adler32}
        if md5:
            dict['md5'] = md5
        if pfn:
            dict['pfn'] = pfn
```

Here `bytes` is the size of the File in bytes, not the actual file data.
Other documentation from the method:
```
        :param scope: The scope of the file.
        :param name: The name of the file.
        :param bytes: The size in bytes.
        :param adler32: adler32 checksum.
        :param pfn: PFN of the file for non deterministic RSE.
        :param md5: md5 checksum.
        :param meta: Metadata attributes.
```

blinkdog:
And finally, the PR is ready for review!

dsschult:
My only question: how much of the Rucio code can be shared for what @vbrik is doing with file replication?

blinkdog:
> My only question: how much of the Rucio code can be shared for what @vbrik is doing with file replication?

`rucio.py` which implements `RucioClient` on top of `RestClient` could probably be reused; the main benefit is `auth()` for handling authentication with Rucio.

Otherwise, it's pretty much down to the nitty gritty of REST calls. The documentation there is poor, and the best way of knowing what to do is reverse engineering a CLI or Python client call, combined with a little trial and error.

PR #33, opened 2019-04-19
use fake token service in tests

dsschult:
I didn't fix up any of @blinkdog's scripts, because I didn't want to go down that rabbit hole.  This was only a 3 point card.

### DIFF
--- a/tests/test_rest_server.py
+++ b/tests/test_rest_server.py
@@ -7,17 +7,28 @@ from math import floor
 import pytest  # type: ignore
 from random import random
 import socket
+import os
 from typing import Dict
 
 from lta.rest_server import boolify, main, start, unique_id
 from pymongo import MongoClient  # type: ignore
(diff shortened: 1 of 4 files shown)


tests/test_rest_server.py
```diff
@@ -36,10 +47,10 @@ def __init__(self, **kwds):
 @pytest.fixture
 def mongo(monkeypatch) -> Database:
     """Get a reference to a test instance of a MongoDB Database."""
-    client = MongoClient("mongodb://localhost:27017/")
+    client = MongoClient(CONFIG['LTA_MONGODB_URL'])
     db = client[MONGODB_NAME]
     for collection in db.list_collection_names():
-        db[collection].delete_many(ALL_DOCUMENTS)
+        db.drop_collection(collection)
```
blinkdog: Good call.

PR #34, opened 2019-06-19
Deleter Module

blinkdog:
Opening a PR so we can do small reviews as the module is coded instead of a big lump at the end.


dsschult:
How does this handle a site like DESY, where we don't want to delete the file, since there is no scratch and we are touching the archive directly?

blinkdog:
We met in my office this morning to talk about Rucio deleting bundles.

Turns out we do want Rucio to manage the archives at some destinations (DESY tape) but not other destinations (NERSC tape).

Also turns out that state in a high level sense (transferring, accessible, deletable) for a transfer request is not easy, as files roll into bundles, bundles can contain files from multiple transfer requests, and these all have their own states which aren't easy to capture at the top level.

PR #36, opened 2019-07-23
Refactored command: ltacmd request new

blinkdog:
Modified lta_cmd.py to POST the correct body to the LTA DB
Modified the LTA DB to build the TransferRequest object correctly
Adjusted the make_token.py helper script to generate scoped roles
Updated the tests for the LTA DB

### DIFF
--- a/lta/lta_cmd.py
+++ b/lta/lta_cmd.py
@@ -100,11 +100,11 @@ async def request_ls(args: Namespace) -> None:
             if not args.long:
                 display_id = display_id[:8]
             create_time = x["create_timestamp"].replace("T", " ")
-            source_cook = x["source"].split(":")
-            path = source_cook[1]
-            source = source_cook[0]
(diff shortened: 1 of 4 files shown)


lta/lta_cmd.py
```diff
@@ -313,7 +315,11 @@ def date_ok(d: str) -> bool:
     # parse the provided command line arguments and call the function
     args = parser.parse_args()
     if hasattr(args, "func"):
-        await args.func(args)
+        try:
+            await args.func(args)
+        except Exception as e:
+            print(e)
+            asyncio.get_event_loop().stop()
```
dsschult: There seems to be a lot of duplicates of `asyncio.get_event_loop().stop()`.  Here, just below, and in the decorator.  Can these be consolidated into one call?

Or maybe we don't want to `loop.run_forever()` in the first place, but rather `asyncio.run(main())`?

PR #39, opened 2019-07-24
Picker Refactoring

blinkdog:
Refactored Picker consumption of TransferRequests from the LTA DB
Added an instance UUID to the Component base class
Modified Picker to provide POST body as LTA DB expects
Refactored LTA DB pop action for TransferRequests


blinkdog:
Note that this refactoring is not complete yet; this is only the first half.
This completed half is the one where the Picker consumes a TransferRequest from the LTA DB.

The half that remains to be finished is the Picker's work:
- Adding Bundles to the LTA DB
- Updating the TransferRequest after its work is finished

lta/rest_server.py
```diff
@@ -767,53 +768,45 @@ class TransferRequestActionsPopHandler(BaseLTAHandler):
...
+            update_doc = {"$set": row}
+            ret2 = await sdtr.find_one_and_update(filter=update_query,
+                                                  update=update_doc,
+                                                  projection=REMOVE_ID,
+                                                  return_document=AFTER)
+            if not ret2:
+                logging.error(f"Unable to claim TransferRequest {uuid}")
+            else:
+                logging.info(f"claimed TransferRequest {uuid} for {claimant}")
+                ret.append(row)
```
dsschult: If we're only ever popping one transfer request, doing a `find` + `find_one_and_update` seems redundant.  Just do the single `find_one_and_update`.

dsschult: We could also not return a list, and just return the single TransferRequest.

tests/test_rest_server.py
```diff
@@ -229,30 +229,36 @@ def test_boolify():
...
+    ret = await r.request('POST', '/TransferRequests/actions/pop?source=WIPAC', wipac_pop_claimant)
     assert ret['results'] == []
 
-    # test limit
+    # test implicit limit
     await r.request('POST', '/TransferRequests', request)
     await r.request('POST', '/TransferRequests', request)
-    ret = await r.request('POST', '/TransferRequests/actions/pop?source=WIPAC&limit=1')
+    ret = await r.request('POST', '/TransferRequests/actions/pop?source=WIPAC', wipac_pop_claimant)
     assert len(ret['results']) == 1
```
dsschult: Do we need the limit test anymore?

lta/rest_server.py
blinkdog: Latest update does it this way.

tests/test_rest_server.py
blinkdog: Test has been removed.

lta/picker.py
```diff
@@ -92,90 +102,71 @@ def _expected_config(self) -> Dict[str, Optional[str]]:
...
+        packing_list = []
+        total_size = 0
+        for catalog_record in catalog_records:
+            file_size = catalog_record["file_size"]
+            #                    0: size    1: full record
+            packing_list.append((file_size, catalog_record))
+            total_size += file_size
+        # divide that by the size requested at the destination
+        bundle_size = self.sites[dest]["bundle_size"]
+        num_bundles = math.floor((float(total_size) / float(bundle_size)) + 0.5)
```
dsschult: Wouldn't `math.ceil(total_size / bundle_size)` be simpler than a floor with +0.5?

Actually, running a test comparing the two, they give slightly different answers.  The current line is actually rounding the answer, and produces identical results to `round(total_size / bundle_size)`.  If this is what we want to do, using the explicit rounding function may be more explicit.

Also note that in python3 you don't need to explicitly cast to float for division.  `/` is float division, while `//` is int division.

tests/test_rest_server.py
```diff
@@ -238,26 +238,20 @@ def test_boolify():
         'claimant': 'testing-picker-aaaed864-0112-4bcf-a069-bb55c12e291d',
     }
     ret = await r.request('POST', '/TransferRequests/actions/pop?source=NERSC', nersc_pop_claimant)
-    assert ret['results'] == []
+    assert ret['transfer_request'] is None
```
dsschult: Instead of asserting None, it might be better to just assert the negation:
`assert not ret['transfer_request']`
This is more future proof in case we decide it should be an empty dict or something else in the future.  Also falls into python's duck typing ethos.

tests/test_rest_server.py
```diff
@@ -238,26 +238,20 @@ def test_boolify():
...
     # repeating gets no work
     ret = await r.request('POST', '/TransferRequests/actions/pop?source=WIPAC', wipac_pop_claimant)
-    assert ret['results'] == []
-
-    # test implicit limit
-    await r.request('POST', '/TransferRequests', request)
-    await r.request('POST', '/TransferRequests', request)
-    ret = await r.request('POST', '/TransferRequests/actions/pop?source=WIPAC', wipac_pop_claimant)
-    assert len(ret['results']) == 1
+    assert ret['transfer_request'] is None
```
dsschult: Same as comment on line 241.

lta/picker.py
blinkdog: They are slightly different in that `round()` breaks ties toward the even number and `math.floor()` after `+0.5` breaks ties toward the next highest integer.
```>>> import math
>>> round(4.5)
4
>>> math.floor(4.5+0.5)
5
```

However, you're right. I agree that `round()` is more explicit; clearly communicating the intent of the code.

Ironically, I tried the `/` division experiment in the Python REPL. It was a Python 2 REPL and it was doing integer division. Having just tried it again in a Python 3 REPL, I confirm that the `float()` casts are not necessary. I've removed them to simplify the code.

The new version is much nicer. Thanks!

tests/test_rest_server.py
blinkdog: Fixed up all 8 of these in `test_rest_server.py`

tests/test_rest_server.py
blinkdog: Fixed up all 8 of these in `test_rest_server.py`

lta/picker.py
dsschult: Round does not in fact always round down:
```
>>> round(5.5)
6
```
The discrepancy is due to floating point precision.  4.5 is probably actually 4.49999...  The pretty printing is getting in the way of how it is at the byte level in IEEE format.

blinkdog: Correct, `round()` rounds up and down, but always breaks ties _toward the even number_:
```
>>> round(1.5)
2
>>> round(2.5)
2
>>> round(3.5)
4
>>> round(4.5)
4
>>> round(5.5)
6
>>> round(6.5)
6
>>> round(7.5)
8
>>> round(8.5)
8
```

dsschult: Interesting.  Strange, but I'm not sure I care enough; and in real life calculations this is a very small edge case we probably don't care about (I want to see the sizes that divide to get exactly .5).

PR #41, opened 2019-08-15
First pass at TransferService refactoring

blinkdog:
`rucio.py` in the root became `transfer/rucio.py` and had the `RucioTransferService` class added.

Most of the code in `RucioTransferService` is pulled out of the Replicator module and reworked a bit to work with the TransferService interface.

The Replicator refactoring is a separate issue, but this should conclude refactoring the Rucio code out to a TransferService class.


lta/transfer/service.py
```diff
@@ -0,0 +1,39 @@
...
+class TransferService:
+    """TransferService is an abstract base class that specifies an interface to transfer services."""
+
+    def __init__(self, config: TransferServiceConfig):
+        """Initialize a TransferService object."""
+        self.config = copy.deepcopy(config)
+
+    async def start(self, spec: TransferSpec) -> TransferReference:
+        """Ask the TransferService to start the specified transfer."""
+        raise NotImplementedError("TransferService.execute() is abstract and must be implemented in a subclass")
```
dsschult: I assume the function name changed.

blinkdog: Good catch. Yes, originally I thought `execute()` but then decided that `start()` more accurately described what the user was actually doing.

PR #42, opened 2019-08-18
First pass at Replicator refactoring

blinkdog:


etc/rucio.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "lta.transfer.rucio.RucioTransferService",
+    "account": "root",
+    "password": "hunter2",
```
dsschult: I have an objection to putting even fake passwords into git.

blinkdog: Move this to an `example` directory and replace the specifics with examples.

PR #43, opened 2019-08-30
First pass at Deleter refactoring

blinkdog:
Still needs:
- Unit tests for Deleter component
- Some refactoring of RucioTransferService

### DIFF
--- a/doc/transfer.md
+++ b/doc/transfer.md
@@ -81,6 +81,7 @@ transfer. The dictionary has the following values:
 
 The `status` field may contain one of the following status codes:
 
+    CANCELED    - Transfer has been canceled (or [cancelled] if it originated in the UK)
     COMPLETED   - Transfer has finished
     CREATED     - Transfer has been created, but not started
     ERROR       - Transfer has stopped due to an error
@@ -154,3 +155,5 @@ The fields have the following meaning:
 * `scope` - The scope to use when creating replicas in Rucio
 * `sites` - Dictionary of the Rucio dataset associated with each site
 * `username` - The username part of the credentials
+
+[cancelled]: https://www.grammarly.com/blog/canceled-vs-cancelled/
--- a/lta/bundler.py
+++ b/lta/bundler.py
@@ -2,7 +2,6 @@
 """Module to implement the Bundler component of the Long Term Archive."""
 
 import asyncio
-from datetime import datetime
 import json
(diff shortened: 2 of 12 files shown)


blinkdog:
Refactored the Deleter's detachment of File DID (replica) from DataSet DID into RucioTransferService.

Still needs:
- `RucioTransferService._force_cleanup()` is not implemented; need to be able to ping Rucio to tell it to physically remove files at the RSE that no longer belong to the given dataset
- `RucioTransferService.status()` has not been implemented yet; this functionality is needed by the Site-Verifier which hasn't been written, and so isn't included in this refactoring issue.

blinkdog:
And now some unit tests round out the refactoring.

Still needs:
- `RucioTransferService._force_cleanup()` implementation
- Integration testing with a live Rucio, to make sure the replicator and deleter work as intended

Ready for review.

blinkdog:
Now I'm not so sure that the Deleter should prod Rucio.
I think the daemon responsible for this is:

https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/reaper/reaper.py#L130

Looking at the code, I don't think there is a simple REST call for this. Rather, the daemon itself uses a Python layer that translates to many REST calls (queries about RSEs, Replicas, etc.) If we attempted to do this with REST calls alone, it would require us to recreate/re-implement the reaper daemon.

Instead this seems like a job for `cron`. Our Rucio server should be running the following every once in awhile: `rucio-reaper --run-once`

I took this command from the documentation here:
https://rucio.readthedocs.io/en/latest/man/rucio-reaper.html

dsschult:
Can I just say "Thanks Rucio for making life confusing?"

If @vbrik agrees with you, I'm ok with this.

vbrik:
rucio-reaper is indeed what rucio uses to delete files.

The cron would probably need to run it with the `--greedy` option if we actually want to delete files. Otherwise, I believe reaper will only delete files if it Rucio believes the RSE to be running out of space (I am not sure how that is determined).

I assume `--rses/--exclude-rses/--include-rses` will also be specified so that we don't delete files from all RSE?

vbrik:
Not sure if this is relevant, but, just in case, note that with `--greedy` option, reaper will delete files that are not part of a replication rule. This could be a problem if you register or upload a file before attaching it to a replication rule.

PR #48, opened 2019-10-11
First pass at NERSC Verifier module

blinkdog:
Just a first pass at the NERSC Verifier module, to allow a Pull Request for review purposes.
Unit tests for coverage have not been written yet.


lta/nersc_verifier.py
```diff
@@ -0,0 +1,240 @@
...
+        # run an hsi command to calculate the checksum of the archive as stored
+        #     hashverify    -> Verify checksum hash for existing HPSS file(s)
+        #     -A            -> enable auto-scheduling of retrievals
+        #     -C            -> cache purge option. If specified, purge files
+        #                      from HPSS disk cache after a successful hash
+        #                      create. Normally used only when it's expected
+        #                      that files will only be fetched once (or infrequently),
+        #                      to help optimize disk cache use
+        #     -H sha512     -> specify that the SHA512 algorithm be used to calculate the checksum
+        args = ["hsi", "hashverify", "-A", "-C", "-H", "sha512", hpss_path]
```
dsschult: :warning: We may want to ask NERSC about this.  If the system is reading the entire tape into cache at once, and we're continually erasing the cache, this could result in poor performance.

lta/nersc_verifier.py
```diff
@@ -0,0 +1,240 @@
...
+        # configure a RestClient to talk to the File Catalog
+        fc_rc = RestClient(self.file_catalog_rest_url,
+                           token=self.file_catalog_rest_token,
+                           timeout=self.work_timeout_seconds,
+                           retries=self.work_retries)
+        # determine the path where the bundle is stored on hpss
+        data_warehouse_path = bundle["path"]
+        basename = os.path.basename(bundle["bundle_path"])
+        stupid_python_path = os.path.sep.join([self.tape_bath_path, data_warehouse_path, basename])
+        hpss_path = os.path.normpath(stupid_python_path)
```
dsschult: We may want to stick something like `tape:` or `archive:` onto the front of `hpss_path`, to make it clear it's not a normal file available for download.

lta/nersc_verifier.py
```diff
@@ -0,0 +1,240 @@
...
+        # create a File Catalog entry for the bundle itself
+        file_record = {
+            "uuid": bundle["uuid"],
+            "logical_name": hpss_path,
+            "checksum": bundle["checksum"],
+            "locations": [
+                {
+                    "site": "NERSC",
+                    "path": hpss_path,
```
dsschult: There should be an `archive` option set here.

lta/nersc_verifier.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        Create a NerscVerifier component.
+
+        config - A dictionary of required configuration values.
+        logger - The object the nersc_verifier should use for logging.
+        """
+        super(NerscVerifier, self).__init__("nersc_verifier", config, logger)
+        self.file_catalog_rest_token = config["FILE_CATALOG_REST_TOKEN"]
+        self.file_catalog_rest_url = config["FILE_CATALOG_REST_URL"]
+        self.tape_bath_path = config["TAPE_BASE_PATH"]
```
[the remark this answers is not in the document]
blinkdog: Wow. Tired eyes miss so much. Corrected this.

lta/nersc_verifier.py
```diff
@@ -0,0 +1,240 @@
...
+                           timeout=self.work_timeout_seconds,
+                           retries=self.work_retries)
+        # determine the path where the bundle is stored on hpss
+        data_warehouse_path = bundle["path"]
+        basename = os.path.basename(bundle["bundle_path"])
+        stupid_python_path = os.path.sep.join([self.tape_bath_path, data_warehouse_path, basename])
+        hpss_path = os.path.normpath(stupid_python_path)
+        # create a File Catalog entry for the bundle itself
+        file_record = {
+            "uuid": bundle["uuid"],
```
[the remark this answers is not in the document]
blinkdog: Hopefully there is not an overlap; that would seem to require a UUID collision, which ought to be extremely rare.

I re-used the UUID here because it seemed weird to have a `logical_name` like `/home/projects/icecube/data/exp/IceCube/2019/filtered/PFFilt/0304/9203b4e0-0cf7-4704-b332-80d15db4d3e5.zip` with a `uuid` field that wasn't `9203b4e0-0cf7-4704-b332-80d15db4d3e5`.

There is something to be said for orthogonal ID spaces; that perhaps allowing LTA to call something by a UUID, and then also have the FC call something else by the same UUID is like playing with fire. I could imagine a scenario where all the jsons are merged into one gigantic omni collection and all queries are always omni queries, then we see where reuse of a "unique" identifier was a bad idea.

However, I was a bit emboldened by the idea that the LTA entity is ephemeral and the FC entity is durable. My thought was the LTA record would disappear so the FC could reuse the UUID without much worry and it would match the filename nicely.

That said, it's just a 'nice to have'; if the ID re-use seems like bad mojo, this is an easy fix.

lta/nersc_verifier.py
```diff
@@ -0,0 +1,240 @@
...
+            # add a location indicating the bundle archive
+            locations = fc_record["locations"]
+            fc_filename = os.path.basename(fc_record["logical_name"])
+            locations.append({
+                "site": "NERSC",
+                "path": f"{hpss_path}:{fc_filename}",
+                "archive": True,
+            })
+            # save the updated entry back to the File Catalog
+            await fc_rc.request("PATCH", f"/api/files/{fc_file_uuid}", locations)
```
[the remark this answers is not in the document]
blinkdog: Ahhh, good catch. I've modified the code to fix this.

I also opened an issue in File Catalog to capture the idea of an API extension:
https://github.com/WIPACrepo/file_catalog/issues/44

I agree with the API extension; lta has the UUID, so a single POST that could save us a GET/PATCH round trip seems nice.

Removing obsolete locations, well ... I'm not sure if there is any help for it.

lta/nersc_verifier.py
blinkdog: Yeah, you're probably right on this. I thought `-C` would be a nice hint to HPSS, like "we're going to calculate the hash, but not otherwise use this file". However, the LTA component has a very narrow scope, and so has limited information to make global decisions like cache eviction. I'll remove it.

lta/nersc_verifier.py
blinkdog: We've had the convention of treating the `logical_name` field like a path. I think adding a tag like that in the path itself will lead to some messy conditional code for handling the data. If we want a tag like that, we could make it another field on the record itself. Maybe `"hpss": true,` ?

lta/nersc_verifier.py
blinkdog: We do use `"archive": True` for files within bundle archives, and give the location a path like:
$BUNDLE_FILE:$FILENAME

The comments in the LTA Design Doc said: `$SITE:$BUNDLE_PATH:$FILENAME` but I thought it was silly to put $SITE in the `"path"` when the `"site"` field right next to it communicated that already (and is the field where FC customers would expect to find that metadata).

That said, if we use `"archive"` to indicate 'file is part of an archive file`, then I don't think we should use the tag on the archive file itself.

lta/nersc_verifier.py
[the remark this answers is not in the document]
dsschult: I guess as long as we are using uuid1 it should be ok.  That's time-based, so unless we're doing a lot of work really close in time it shouldn't matter.

Also, I guess the FC will reject the request if there's an overlap.

lta/nersc_verifier.py
dsschult: Maybe just `"online": false` to signal that this filesystem is not for online access?

dsschult: This is very similar to the archive flag in meaning though.  This is a tossup as to what to use.

lta/nersc_verifier.py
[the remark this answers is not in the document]
blinkdog: Still need the logical_name in order to specify the location.
There is no beating the round-trip it seems.
However, using the POST route feels more correct, and once the FC is refactored should be a lot safer than a PATCH overwrite race.

lta/nersc_verifier.py
blinkdog: I put both `"hpss": true,` and `"online": false,` into the location for the archive bundle.

PR #56, opened 2019-12-03
Check and Load LTA

blinkdog:
Refactored ltacmd; added new `ltacmd catalog check --path $PATH` command.

This PR isn't finished yet, just opening it up for review purposes.


dsschult:
My only comment is a question on whether this should actually be a part of `ltacmd`, or broken out separately.  Most of the checking is between the FC and the PATH on disk, which doesn't involve LTA at all.

PR #60, opened 2019-12-12
Modified the order of bundling steps to accommodate JADE-LTA

blinkdog:


lta/bundler.py
```diff
@@ -176,38 +177,43 @@ def _check_mysql(self) -> bool:
...
         self.logger.info(f"Adding row to MySQL Database: {self.user}@{self.host}/{self.db}")
         self._insert_jade_row(bundle)
+        # 8. Move the bundle from the work box to the outbox
+        if final_bundle_path != bundle_file_path:
+            self.logger.info(f"Moving bundle from '{bundle_file_path}' to '{final_bundle_path}'")
+            shutil.move(bundle_file_path, final_bundle_path)
+        self.logger.info(f"Finished archive bundle now located at: '{final_bundle_path}'")
+        # 9. Update the Bundle record in the LTA DB
+        self.logger.info(f"PATCH /Bundles/{bundle_id} - '{bundle}'")
+        await lta_rc.request('PATCH', f'/Bundles/{bundle_id}', bundle)
```
dsschult: So for JADE LTA it's required that the DB row is created before the file is moved?

blinkdog: Not strictly so; it's possible to adjust the glue.

However, the JADE pattern is very much file handoff is ownership handoff. Doing work after moving the file to the outbox invites race conditions.

In LTA, we do it the opposite way. We're allowed to move the file around as much as we want, up until we do the handoff in the REST server. Doing work after updating the REST server would invite the race condition.

dsschult: Ah, OK.

PR #72, opened 2020-01-16
Add preflight checks to NerscMover work cycle

blinkdog:


dsschult:
About the only difference of opinion I have is about the "too many nersc movers" issue.  Because you have to submit them to a batch system at NERSC, I would assume whatever script submits the batch jobs could just not queue more than N jobs.

But maybe I'm missing something.

jnbellinger:
We can run no more than 14 jobs that touch the HPSS.
We can run up to 100 slurm jobs
We can run  some finite number of cron-ized nersc_mover jobs.  These can't be daemons, since user jobs can't run more than 48 hours on a cori system--they've got to be one-shots or "semi-daemons".  The model here seems to be that 1 nersc_mover makes 1 or fewer connections to the HPSS at a time, so yes, we can control the number of connections by controlling the number of nersc_mover jobs.
But what happens when we want to pull as well as push?  The simplest approach seems to be to have the REST server count up how many connections are being made, and make sure that the nersc_moverX jobs check if they are green-lighted.

dsschult:
Maybe I need to see on a whiteboard exactly what you're planning for NERSC.

PR #74, opened 2020-01-22
Added script to schedule NERSC components as SLURM jobs

blinkdog:


blinkdog:
The script was written by Jim Bellinger.
I have copied it into the LTA code tree and created this pull request to allow review.

dsschult:
My comment at first glance is that I hate reading bash scripts that are non-trivial :smile:

jnbellinger:
I understand very well.  I'm not always keen on reading my own after they've aged a year or two.
This is muddier than I wanted it to be because I wanted to include some local testing before turning it loose on an unsuspecting NERSC.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,273 @@
...
+   function SQUEUE {
+      base="CLUSTER: escori\n        JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)\n"
+      base=${base}"\n"${NS_FAKE_QUEUE}
+      echo "${base}"
+      return ${NS_FAIL_TEST_SQUEUE}
+   }
+ else
+   export NS_LOG=/global/homes/i/icecubed/LTA/ns.log
+   export NS_SLURM_LOG=/global/homes/i/icecubed/LTA/SLURMLOGS
+   export NS_SLURM_LOG_SEEN=/global/homes/i/icecubed/LTA/SLURMLOGS/seen
```
barnetspa: For config variables like this, the parent path should be declared as a variable and then used here: 

WORKING_DIR=/global/homes/i/icecubed/LTA
NS_LOG=$WORKING_DIR/ns.log

etc

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,273 @@
...
+         fi
+    done
+  echo "${unfinished}"
+  return 0
+}
+
+###
+# Check the slurm job activity information, count the incomplete entries
+
+function countIncomplete {
```
barnetspa: This code should be moved to a standalone check?

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,273 @@
...
+###
+# Inspect the slurm logs
+# If completed, move to "seen"
+# If not completed, count and make that available to the main
+#  program for comparison w/ the number of slurm jobs.  We
+#  can have a race condition in which one of the incomplete
+#  slurm jobs completes by the time we ask what's active.
+# I figure the process itself (and its heartbeat) will manage
+#  the error reporting
+function cleanupLogs {
```
barnetspa: This code is to be moved to a separate process.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,273 @@
...
+  if [[ ! -f ${NS_SCRIPT_PATH}/$1".sh" ]]
+    then
+      logit 0 "Submit script $1.sh was not found"
+      return 1
+    fi
+  logit 2 "launch $1"
+  if ! SBATCH  --comments="${1}" -o "${NS_SLURM_LOG}/slurm-$1-%j.out" "${NS_SCRIPT_PATH}/$1.sh"
+    then
+      logit 0 "Submit of $1 failed"
+      return 1
```
barnetspa: Return codes should be different for these errors.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,273 @@
...
+#
+declare -i NS_LOG_DETAIL count expectedcount fc incomplete activeJobs
+
+export NS_DEBUG="True"
+
+###############################
+# Setup for debugging/testing #
+###############################
+
+if [[ ${NS_DEBUG} == "True" ]]
```
barnetspa: As this code is re-factored, things like log level and debug flags should be moved to either config or command line (or both).

blinkdog:
This Pull Request has been re-imagined at:
https://github.com/WIPACrepo/lta/pull/96

PR #94, opened 2020-02-07
enable mongodb index profiling

dsschult:
It looks like there are a lot of indexes needed in LTA.


dsschult:
Apparently I didn't start from an updated master.  Will fix.

dsschult:
ok, the test is now failing appropriately, highlighting non-indexed queries.

dsschult:
@blinkdog everything look good now?

PR #96, opened 2020-02-10
Adding nersc_supervisor.sh

jnbellinger:
How does this look?


resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+# Ask sacct about the job history, and figure out how many jobs have not completed.
+#  This may be for a variety of reasons, listed below
+# Ask squeue for which jobs are currently running.  If jobs have failed they
+#  won't appear here, but will appear in the previous list.
+#
+# The second method is definitive for determining whether or not to
+#  submit a new job.  However, the sacct information tells me if something
+#  needs to be looked at.
+#
+# The sacct info could be accumulated with a different script, if desired.
```
blinkdog: I think this would be the right way to go.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+
+###
+# Loop over the list of expected job types
+# At the moment this info is hard-wired in NS_DESIRED, but
+# we should load from a json initialization file instead
+# Keeping track of activeJobs is future-proofing the system
+# I should count them all and compare with maxjobs before
+# submitting anything.
+###
+activeJobs=0
```
blinkdog: It looks like the code below is counting up the jobs from those specified in `NS_DESIRED`.
However, if there were other jobs running (of a kind not specified in `NS_DESIRED`) it looks like it wouldn't be counted as part of `activeJobs`.
Since `activeJobs` is just future-proofing, there isn't really anything to change here yet.
Just something (non-specified jobs) to keep in mind when building that future proofing at a later date.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+        if ! launch "${class}"
+          then exit 3; fi
+        logit 2 "Launching ${class} ${count} ${expectedcount}"
+      fi
+  done
+
+########
+# Done #
+########
+exit 0
```
blinkdog: On line 189 you've got a logit for invocation.
Do you also want one here at the end for normal termination?

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+###############
+#
+#  The slurm command options have not changed since 22-Jan-2020
+#
+#  The names of the scripts are as in 22-Jan-2020, with
+# a XXX.sh script which defines the environment and invokes
+# a XXX.py script which does the phone-home etc.
+#
+#  The scripts to invoke are in the directory ABOVE the
+# directory containing this script.
```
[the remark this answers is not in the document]
jnbellinger: Yep.  Cleaning up the comment.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+
+declare -i NS_LOG_DETAIL count expectedcount fc activeJobs
+#declare -i incomplete
+
+export NS_BASE=/global/homes/i/icecubed/NEWLTA/lta
+export NS_LOG="${NS_BASE}/ns.log"
+export NS_SLURM_LOG="${NS_BASE}/SLURMLOGS"
+export NS_SLURM_LOG_SEEN="${NS_SLURM_LOG}/seen"
+NS_SCRIPT_PATH="${NS_BASE}/bin"
+export NS_SCRIPT_PATH
```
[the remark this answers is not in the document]
jnbellinger: Sure.

resources/nersc_supervisor.sh
jnbellinger: Good point

resources/nersc_supervisor.sh
jnbellinger: Yes, I'll have to have a full list of what's expected--because slurm jobs may not be HPSS jobs.

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+
+##################
+# Start the work #
+##################
+
+#if ! incomplete=$(countIncomplete)
+#  then
+#    exit 1
+#  fi
+
```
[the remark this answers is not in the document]
jnbellinger: Trimmed

resources/nersc_supervisor.sh
```diff
@@ -0,0 +1,248 @@
...
+    then
+        echo 0
+        return 0
+    fi
+  # Steve doesn't think I need to log the incomplete info
+  #logit 2 "${listing}"
+  # Count the number of not-completed, skipping the terminal "|" if any
+  echo "${listing}" | /usr/bin/awk '{n=split($0,a,"|");m=n-1;if(m<0)m=0;print m;}'
+  return 0
+}
```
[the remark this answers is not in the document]
jnbellinger: Trimmed

PR #97, opened 2020-02-10
Change ltacmd file catalog lookup to use logical_name

blinkdog:


blinkdog:
https://wipacit.atlassian.net/browse/DEV-837

PR #114, opened 2020-04-02
Deleter is now RucioDetacher

blinkdog:


blinkdog:
https://wipacit.atlassian.net/browse/DEV-880

blinkdog:
I called it the RucioDetacher because that's how Rucio refers to the process of disconnecting one DID from another:

https://rucio.readthedocs.io/en/latest/restapi/did.html#delete--dids-(scope)-(name)-dids

"Detach data identifiers from data identifiers."

Because this is the REST call underlying the process, I thought it was best to (re-)use Rucio terminology.

PR #136, opened 2020-09-15
Modify lta_cmd to use lazy initialization for commands

blinkdog:


dsschult (changes requested):
This should be a lot simpler than injecting the dependencies in every function.  How about just filling in `args` before you call `args.func(args)`?  At that point we know they're not asking for help.

lta/lta_cmd.py
```diff
@@ -57,6 +57,19 @@
     "/data/sim",
 ]
 
+# -----------------------------------------------------------------------------
+
+di: Dict[str, Any] = {}
```
dsschult, line 62: Can we not use a global variable?  This feels wrong.

dsschult (approved):
I mostly like this now.  Not sure that `di` is the best name for config, but that's a minor detail.

PR #137, opened 2020-09-18
Use fixed size bundles and limit bundle file count

blinkdog:
https://wipacit.atlassian.net/browse/DEV-1005 : LTA Picker should set firm bundle size limits
https://wipacit.atlassian.net/browse/DEV-1001 : LTA Picker split by File Count as well as File Size


dsschult (approved):
I looked that helper function up.  Useful for this.

PR #150, opened 2020-12-28
Improve GET /Bundles query performance

blinkdog:
* GET /Bundles now requests projection of `uuid` to save traffic between LTA DB and Mongo
* GET /Bundles/{uuid} has new query parameter `contents` that defaults to True, by adding `?contents=0` to queries, the huge field `files` can be excluded from the result returned by the LTA DB.
* `ltacmd` script updated to make use of the performance gains when bundle contents are not needed
* Version bumped, Changelog updated, Requirements updated


jnbellinger (approved):
Offline explanation was appreciated.

PR #172, opened 2021-02-23
File Metadata lives in LTA DB not a Bundle spec

blinkdog:
This PR is a work in progress:
- [x] LTA DB
- [x] ltacmd
- [x] picker
- [x] bundler
- [x] nersc_verifier
- [x] desy_verifier
- [x] locator
- [x] unpacker

File metadata used to live in the Bundle record as full copies of the File Catalog entries.
Later, these file metadata records were trimmed down to a small set of keys (uuid, checksum, etc.)
Now, we remove the file metadata from the Bundle record altogether.

When Picking files to bundle, we add mapping records to a new LTA collection `Metadata` (#177):
```
{
    "uuid": $METADATA_UUID,
    "bundle_uuid": $BUNDLE_UUID,
    "file_catalog_uuid": $FILE_CATALOG_UUID,
}
```

This has two effects:
1. Bundle records are not bogged down the File metadata. We can exceed the previous limit of 25K files per Bundle because now we don't pay per-file in the Bundle. A million files? Fine.
2. We can query the LTA DB for the files associated with a Bundle, and process them in small (~1000 record) batches.


dsschult:
I'm unsure it's a good idea to pollute the File Catalog with temporary LTA metadata.  This is also "yet another thing" we have to index, otherwise the query could take an hour to return the files in a bundle (doing a full collection scan).

Feel free to persuade me to your point of view though.

jnbellinger:
What kind of performance hit might we take scanning all the files to see which have the bundle uuid(s) associated?  There could easily be more than one associated bundle, e.g. in the two cases of 1) pilot error double-bundles the directory (mea culpa) and 2) bundles made at different times that go to different sites.

blinkdog:
> I'm unsure it's a good idea to pollute the File Catalog with temporary LTA metadata. This is also "yet another thing" we have to index, otherwise the query could take an hour to return the files in a bundle (doing a full collection scan).
> 
> Feel free to persuade me to your point of view though.

Well, the File Catalog has always had room for application-specific metadata under it's own key, and the LTA metadata is relatively small (a uuid and a date).

Originally, I thought to do another collection in LTA itself, add a `/Metadata/${BUNDLE_UUID}` route to the REST interface and collect the files there. In this scenario, we still have an index, it's just in the LTA DB database instead of the File Catalog database. We also have another collection, whose records are (bundle_uuid, file_catalog_uuid) tuples; add in date or other metadata too as desired.

It occurred to me that the File Catalog already has records for the files, already has routes for querying and pagination, we'll need an index any way we slice it, and we'll be modifying tons of File Catalog records anyway when the bundle gets verified (to add the new location). So why not tag the record we intend to update later anyway?

It is a little less pure. LTA is storing some of its metadata in the File Catalog instead of its own database. If that's a deal breaker we can pull it over to the LTA side. It might take a little longer because we need to implement the /Metadata route(s), but that seems like ~1 week, so nothing terrible.

blinkdog:
> What kind of performance hit might we take scanning all the files to see which have the bundle uuid(s) associated? There could easily be more than one associated bundle, e.g. in the two cases of 1) pilot error double-bundles the directory (mea culpa) and 2) bundles made at different times that go to different sites.

True. If we go with metadata on the file catalog record, it could probably use a little more robust schema.
One thing to check would be if our current version of Mongo can index array content.

jnbellinger:
After sleeping on it:  
One of the use cases for the F.C. is to locate data:  e.g. you learn of a hidden supernova and want to fetch a couple month's worth of PFRaw from _wherever_ they happen to be.  So with each file are pointers to something location-y:  in this case a bundle id and, associated with that, a location: NERSC_Bundles:/home/projects/icecube/data/exp/IceCube/2012/unbiased/PFRaw/0230/.  It could also be a disk UUID and a disk location:  ChamberlinDiskArchive_Season10_Disk312:/data/exp/etc , but maybe we don't want the hassle of combining the Pole db with the filecatalog.

Anyhow, unless I'm still groggy, it seems as though the file record will want to keep around pointers that might look something like what you suggest, although maybe the details of pending/picking/completeFileList go elsewhere.

So, if as part of LTA I want to know if a file has been bundled, I query the file for its "bundle" pointers, then query the "bundle" entry for its location and status.  That's 2 queries (I could maybe get smart with grouping them) but it keeps the info I have to stuff into a given file record to a minimum, and it seems like logical stuff.  The "bundle" table holds the status, dates, location, and pointers to the files it contains.  It doesn't seem like "pollution", though the dates and error messages will become obsolete as soon as LTA is done with it.

dsschult:
@jnbellinger those "pointers" are the location data we already record, with "site", "path", and "archive" keys.  We already update that for each file when the bundle is verified.

So it comes down to where to store the more temporary LTA data.

dsschult:
> > What kind of performance hit might we take scanning all the files to see which have the bundle uuid(s) associated? There could easily be more than one associated bundle, e.g. in the two cases of 1) pilot error double-bundles the directory (mea culpa) and 2) bundles made at different times that go to different sites.
> 
> True. If we go with metadata on the file catalog record, it could probably use a little more robust schema.
> One thing to check would be if our current version of Mongo can index array content.

Yes: https://docs.mongodb.com/manual/core/index-multikey/

So you just need to format the metadata correctly to hit an index.

blinkdog:
> I'm unsure it's a good idea to pollute the File Catalog with temporary LTA metadata. This is also "yet another thing" we have to index, otherwise the query could take an hour to return the files in a bundle (doing a full collection scan).
> 
> Feel free to persuade me to your point of view though.

When switching gears to fixing up the pipeline, I realized that having things scattered about is one of my pain points for LTA.
Now that I think about it, it's more convenient programming wise to tag the files in the File Catalog, but it's more convenient mentally to know all the temporary metadata lives in a single place in the LTA DB.

I think I'd rather do the `/Metadata` routes after all.

PR #174, opened 2021-02-24
Update base image of Dockerfile

blinkdog:


dsschult (approved):
My only question is if upgrading pip is necessary, since I'd assume the latest build will have a pretty recent version (and any security updates).

PR #192, opened 2021-03-30
make docker tags for releases

dsschult:


blinkdog (approved):
LGTM

The `Compute Docker Tag` step on L15-16 doesn't look like it's necessary any more. Removing L15-17 would fix it right up.

However, that's not a deal-breaker. The step is a bash one-liner, so if it stays in :shrug: it's not going to hurt anything.

dsschult:
I removed that step.

Also did some cleanup based on things I learned with the File Catalog.  Combining the two files into one, with a conditional on pushing to dockerhub.  And only tagging the semantic versions, not the raw release name.

blinkdog (approved):
Very nice! Saves duplicating the Docker action changes across two actions.

PR #196, opened 2021-04-19
NERSC's HPSS has binaries in new locations

jnbellinger:


lta/nersc_mover.py
```diff
@@ -2,6 +2,7 @@
 """Module to implement the NerscMover component of the Long Term Archive."""
 
 import asyncio
+from logging import Logger
```
blinkdog, line 5: Our integration with LGTM doesn't like that we use `from X import Y` and `import X`.
Could you remove this import and restore `Logger = logging.Logger` below?

lta/nersc_mover.py
```diff
@@ -15,7 +16,6 @@
 from .log_format import StructuredFormatter
 from .lta_types import BundleType
 
-Logger = logging.Logger
```
blinkdog, line 18: This is the `Logger = logging.Logger` that I was talking about.

lta/nersc_retriever.py
```diff
@@ -2,6 +2,7 @@
 """Module to implement the NerscRetriever component of the Long Term Archive."""
 
 import asyncio
+from logging import Logger
```
blinkdog, line 5: Same thing here, need to fix it for LGTM.

lta/nersc_verifier.py
```diff
@@ -2,6 +2,7 @@
 """Module to implement the NerscVerifier component of the Long Term Archive."""
 
 import asyncio
+from logging import Logger
```
blinkdog, line 5: Same thing here, need to fix it for LGTM.

blinkdog (changes requested):
The path changes for the NERSC binaries looks good.

Plumbing them through configuration may allow more flexibility for the future, but that's more complicated and higher touch, so no need to do it now. If NERSC changes up the paths again before 2022, we should consider it though.

LGTM will complain about the import change, so if we can fix those up, I think this one is ready to merge.

blinkdog (changes requested):
Looks like the path for `hsi` is incorrect.

```
[pmeade@dtn01 ~]$ ls /usr/bin/hsi
ls: cannot access /usr/bin/hsi: No such file or directory
[pmeade@dtn01 ~]$ ls /usr/local/bin/hsi
/usr/local/bin/hsi
```

jnbellinger:
icecubed@cori08:~/NEWLTA/lta> which hsi
/usr/bin/hsi

It's very odd that dtn systems should be different, but these scripts run on the cori systems

blinkdog:
> icecubed@cori08:~/NEWLTA/lta> which hsi
> /usr/bin/hsi
> 
> It's very odd that dtn systems should be different, but these scripts run on the cori systems

Wow, that is surprising! Never mind then, carry on with `/usr/bin/hsi` :smile:

blinkdog:
@jnbellinger If you fix the LGTM issues with the `import` and `import from`, this one is ready to merge.

PR #197, opened 2021-04-23
Unpacker can rewrite physical paths to logical paths

blinkdog:


ric-evans:
@blinkdog what are you using for `PATH_MAP_JSON`? Is this future-proofed for future lfs versions?

blinkdog:
> @blinkdog what are you using for `PATH_MAP_JSON`? Is this future-proofed for future lfs versions?

`PATH_MAP_JSON` is a path to a file containing JSON.
The JSON is a map, like this:

```json
{
    "/mnt/lfs7": "/data/exp"
}
```

When the Unpacker finds paths that begin with `/mnt/lfs7` they will be replaced with `/data/exp`.
I'm not sure if this is 100% future-proof, the JSON may need to be updated with new mappings if the need arises.

Best practice would be to archive by logical path (`/data/exp...`) and avoid all physical paths (e.g. lfs) in the future.

Does this answer the question, @ric-evans ?

ric-evans:
Yeah, that's what I was wondering.

> Best practice would be to archive by logical path (/data/exp...) and avoid all physical paths (e.g. lfs) in the future.
The Indexer does that. If anything other than `/data/`-rooted paths are passed, it fails loudly.

PR #219, opened 2021-11-12
Minimum needed for running site-move-verifier at DESY.  Intended to r…

jnbellinger:
…eplace pipe2-desy-move-verifier


blinkdog (approved):
LGTM.

The other pipelines don't seem to use `use_full_bundle_path = True` on this component so it looks like it's okay to modify the logic there.

PR #220, opened 2021-11-16
Trim down some async warnings

blinkdog:
As discussed with David and Ric on Slack the other day, there are a few warnings we can trim down in the tests.

`AsyncMockMixin._execute_mock_call` is a bit more tricky.

https://bugs.python.org/issue40406

This PR also replaces a home-rolled `AsyncMock` with the official `AsyncMock` from Python 3.8

### DIFF
--- a/tests/test_bundler.py
+++ b/tests/test_bundler.py
@@ -2,14 +2,13 @@
 """Unit tests for lta/bundler.py."""
 
 import os
-from unittest.mock import call, mock_open, patch
+from unittest.mock import AsyncMock, call, mock_open, patch
 from uuid import uuid1
 
 import pytest  # type: ignore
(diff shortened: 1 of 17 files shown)


blinkdog:
> `AsyncMockMixin._execute_mock_call` is a bit more tricky.
> https://bugs.python.org/issue40406

Just a note to myself...

This is an AsyncMock having a mocked method called without `await`.
If `AsyncMock` is standing in for some object, it may offer non-async methods; in this case they should be explicitly populated with a `MagicMock` that doesn't need to be `await`ed.

blinkdog:
It's not perfect, but it does fix some warnings and I'd like to close it out and get it off the list.

Issue #222, opened 2021-12-09
rate-limit step picked up the wrong bundle name


jnbellinger:
A bundle was created, and sat in bundler_work
The rate-limiter (stager) claimed the bundle, and was given a bundle zip name that did not match the bundle uuid.  It couldn't find the bundle, of course, and failed.
A ./ltacmd status returned the right information, and resetting the status and letting the system claim it again worked fine.
So we have an intermittent problem

Example:

Bundle **ad83f1b04f1211ecb432a2d644f9f2f8**
  Priority: 2021-12-04 00:35:36
  Status: quarantined (2021-12-04 00:35:36)
    Reason: BY:n-24.icecube.wisc.edu-pipe2-rate-limiter-7fff7570-03e7-47ba-93ae-b10d2e201d5d REASON:[Errno 2] No such file or directory: '/mnt/lfs7/ltatemp/bundler_todesy/**74836edc4f1111ecb432a2d644f9f2f8.zip**'
  Claimed: True
    Claimant: n-24.icecube.wisc.edu-pipe2-rate-limiter-7fff7570-03e7-47ba-93ae-b10d2e201d5d (2021-12-04 00:35:36)
  TransferRequest: 9db0e1304f1211ecb432a2d644f9f2f8
  Source: WIPAC -> Dest: DESY
  Path: /data/exp/IceCube/2019/filtered/PFFilt/0527
  Files: Not Listed
  Bundle File: /mnt/lfs7/ltatemp/bundler_stage/ad83f1b04f1211ecb432a2d644f9f2f8.zip

jnbellinger:
The rate limiter is run every 15 minutes, and is on a run-once-and-die schedule.  I demand that 1 exist, so it should not launch a second--but that is the sort of failure mode it looks like.

The fact that it was processed is not indicative of anything, unfortunately--I had restarted it to see what would happen, and it ran correctly.


It happened again:
c6cd775a5c3b11ecb432a2d644f9f2f8
tried to find 2e8de5f25c3a11ecb432a2d644f9f2f8.zip in bundler_todesy/
instead of
c6cd775a5c3b11ecb432a2d644f9f2f8.zip
in bundler_stage/

The c6cd7... bundle was quarantined at 9:35, and the 2e8de... bundle was declared "finished" at 9:36, so it must have been deleted just before then.

I'm not sure what sort of inter-module collision would cause this, but several different ones are spawned within a second or so.

jnbellinger:
[...]

PR #224, opened 2021-12-16
More logging for rate_limiter

blinkdog:
Will help to clarify the root cause of #222


jnbellinger (commented):
This looks like the area that needs inspection.  Fine by me

PR #226, opened 2021-12-22
Allow rate-limiter to ignore missing bundles in the destination check

### LINKED ISSUE #222: rate-limit step picked up the wrong bundle name

blinkdog:
This modifies the rate-limiter component to retry when bundles in the destination directory are removed when the rate-limiter is in the middle of counting and sizing the files in the destination directory.

Fixes #222

### DIFF
--- a/lta/rate_limiter.py
+++ b/lta/rate_limiter.py
@@ -81,9 +81,9 @@ class RateLimiter(Component):
             try:
                 # determine the size of the file
                 size += os.path.getsize(disk_file)
-            except:
+            except Exception as e:
                 # whoops, looks like somebody downstream moved it
-                self.logger.error(f"Skipped getsize() on missing file: {disk_file}")
+                self.logger.error(f"Skipped getsize() on missing file: {disk_file}", exc_info=e)
                 continue
             self.logger.debug(f"Size so far: {size} bytes")
         self.logger.info(f"Found {len(disk_files)} entries ({size} bytes) in {path}")
--- a/lta/rate_limiter.py
+++ b/lta/rate_limiter.py
@@ -78,23 +78,16 @@ class RateLimiter(Component):
         self.logger.info(f"Determining total size of files in {path}")
         size = 0
         for disk_file in disk_files:
-            # determine the size of the file
-            size += os.path.getsize(disk_file)
+            try:
+                # determine the size of the file
+                size += os.path.getsize(disk_file)
+            except:
+                # whoops, looks like somebody downstream moved it
+                self.logger.error(f"Skipped getsize() on missing file: {disk_file}")
+                continue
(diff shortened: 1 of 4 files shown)


lta/rate_limiter.py
```diff
@@ -84,6 +84,18 @@ def _get_files_and_size(self, path: str) -> Tuple[List[str], int]:
         self.logger.info(f"Found {len(disk_files)} entries ({size} bytes) in {path}")
         return (disk_files, size)
 
+    def _get_files_and_size_with_retry(self, path: str) -> Tuple[List[str], int]:
+        """Recursively walk and add the files of files in the file system."""
+        retry_left = 3
```
jnbellinger, line 89: I can think of two causes for a failure:  
1) The file isn't there any more--which is a good thing, and we should just skip on to the next bundle
2) The filesystem is having issues--a bad thing, which will eventually manifest as a failure to move the bundle from staging to output
The former doesn't need to raise an exception, and the latter will do so later.  I'd think a simple try/except around the filesystem accesses on lines 64 and 82 would be better than raising an exception when the file is missing.

jnbellinger: Something like:
        for disk_file in disk_files:
            # determine the size of the file
            try:
                size += os.path.getsize(disk_file)
            except:
                # File gone already?  OK.
                continue
            self.logger.debug(f"Size so far: {size} bytes")

PR #259, opened 2023-05-30
bump py-versions CI release v2.1

alemsh:


alemsh:
also updated try-setup-install to use the py-version action instead of the local script

PR #269, opened 2023-10-05
Update scrontab for prune slurm logs

blinkdog:
`cd /global/homes/i/icecubed/lta/slurm-logs; find . -type f -mtime 7 -delete`

Not the most brilliant construction. If `/global/homes/i/icecubed/lta/slurm-logs` doesn't exist or is somehow inaccessible, it deletes things exactly a week old wherever scrontab jobs happen to start ... is that `${HOME}`? No idea.

Replaced with:
`find /global/homes/i/icecubed/lta/slurm-logs -type f -mtime +7 -delete`

This will find in the provided directory, or fail, and the `+7` means we get all the logs one week and older, not just those exactly a week old.

### DIFF
--- a/nersc/scrontab
+++ b/nersc/scrontab
@@ -33,4 +33,4 @@
 #SCRON -t 00:15:00
 #SCRON -o /global/homes/i/icecubed/lta/slurm-clean.out
 #SCRON --open-mode=append
-0 2 * * * cd /global/homes/i/icecubed/lta/slurm-logs; find . -type f -mtime 7 -delete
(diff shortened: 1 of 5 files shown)


blinkdog:
This is failing on test setups that want new Python 3.12 dependencies, but...

1. We use Python 3.10 in the deployment on Perlmutter at NERSC.
2. This PR is not a code change; it captures a configuration change already made at NERSC.

I'll merge this one and hope for better 3.12 support in CI/CD the future.

PR #273, opened 2024-01-11
fix tests

dsschult:
Needed to remove python 3.8, as the new ArgumentParser does not support it.


lta/rest_server.py
```diff
@@ -766,11 +782,18 @@ def start(debug: bool = False) -> RestServer:
...
+    else:
+        auth = {
             "audience": config["LTA_AUTH_AUDIENCE"],
-            "openid_url": config["LTA_AUTH_OPENID_URL"],
-        },
+            "openid_url": config["LTA_AUTH_OPENID_URL"]
+        }
+
+    args = RestHandlerSetup({  # type: ignore
+        "auth": auth,
```
blinkdog, line 796: This is really elegant; much nicer than the hacky global scope lta_auth swap. Thank you for this.

blinkdog (approved):
LGTM

PR #274, opened 2024-01-18
Fix Paths and Pins

blinkdog:
* Modifies `collabssh` so that it invokes `sshproxy.sh` without requiring that the current directory be on one's PATH.
* Adds a lower bound on the `wipac-rest-tools` requirement; use of `ArgumentHandler` requires >=1.6.0
    * Pip was choosing from the 1.5.x series, which will no longer work.

### DIFF
--- a/collabssh
+++ b/collabssh
@@ -4,5 +4,5 @@
 # Use direct login to ssh to perlmutter with the icecubed collaboration account
 # See: https://docs.nersc.gov/accounts/collaboration_accounts/#direct-login
 
-sshproxy.sh -c icecubed
+./sshproxy.sh -c icecubed
 ssh -i ${HOME}/.ssh/icecubed icecubed@perlmutter-p1.nersc.gov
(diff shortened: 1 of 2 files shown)


dsschult:
It would be interesting to see the log from pip where it was choosing the 1.5.x series, and why it was doing so.

dsschult (approved):

blinkdog:
[pip-install-output.txt](https://github.com/WIPACrepo/lta/files/13992596/pip-install-output.txt)

It's interesting because it picks 1.6.0 at first, then after requests-futures it looks for compatible (~1.5.x) versions:

```
Collecting requests-futures (from wipac-rest-tools->lta==0.42.0)
  Downloading requests_futures-1.0.1-py2.py3-none-any.whl.metadata (12 kB)
INFO: pip is looking at multiple versions of wipac-rest-tools to determine which version is compatible with other requirements. This could take a while.
Collecting wipac-rest-tools (from lta==0.42.0)
  Downloading wipac_rest_tools-1.5.3-py3-none-any.whl.metadata (6.1 kB)
  Downloading wipac_rest_tools-1.5.2-py3-none-any.whl.metadata (6.1 kB)
```

PR #275, opened 2024-02-24
update to more modern tornado asyncio wait syntax

dsschult:
Fixes server not responding to requests error in prod.


blinkdog (approved):
Thanks for quick fix.

I added a commit to fix up the test that was freezing, and a little housekeeping on some support scripts.

I'll go ahead and merge it now.

PR #279, opened 2024-04-22
Require rest-tools that supports leeway

blinkdog:
David provided a PR (https://github.com/WIPACrepo/rest-tools/pull/142) that set a default leeway time for JWT auth in rest-tools.
This configuration change ensures that LTA makes use of this new leeway for authentication.


blinkdog:
Looks like we had the same idea #280

blinkdog:
Covered in #280

PR #280, opened 2024-04-22
bump wipac-rest-tools

dsschult:


lta/__init__.py
```diff
@@ -9,7 +9,7 @@
 # is zero for an official release, positive for a development branch,
 # or negative for a release candidate or beta (after the base version
 # number has been incremented)
-__version__ = "0.42.0"
+__version__ = "0.41.33"
```
blinkdog, line 12: Waaaa? :question:

blinkdog (changes requested):
This is a little strange...

lta/__init__.py
dsschult: So the semantic release tool really didn't like that the master branch code version was ahead of the github tag number.  They're supposed to be equal I guess?

blinkdog: Okay. I guess it doesn't matter so much, it just seemed weird for the version to decrease.

Want to bring my change from #279 in to this PR, and we can close mine and merge this one?

blinkdog: :facepalm: I missed it the first time.

blinkdog (approved):
Let's go! :smile:

PR #281, opened 2024-07-09
Reduce Authentication Requests to Keycloak

blinkdog:
@vbrik reached out on Slack because LTA was making a lot of requests to the Keycloak service.

Some of this is unavoidable; there are 10 components in a transfer pipeline, and all of them need to talk to the LTA REST service in order to coordinate their work.

Also, at peak times (retro disk processing) there can be 16-32 instances of some components like bundler and gridftp_replicator.

NERSC's slurm based job system is not great for long-lived components, so depending on how aggressively these components at scheduled at NERSC, they will start up frequently and authenticate with keycloak every time.

However, there are some improvements that can be made to the code. Instead of authenticating with keycloak at every work cycle, the RestClient can be created when the component is started, and the client can re-use its existing token, and only refresh it when necessary.

This PR introduces the code changes needed to create the RestClient early and provide it to the component to enable re-use.

A similar issue may affect the RestClient the component uses to access the File Catalog. Fortunately, these requests to authenticate with Keycloak should only happen when actual work is being performed. If these are also very noisy, we may consider earlier creation of this resource as well.

### DIFF
--- a/lta/__init__.py
+++ b/lta/__init__.py
@@ -9,7 +9,7 @@ from __future__ import absolute_import, division, print_function
 # is zero for an official release, positive for a development branch,
 # or negative for a release candidate or beta (after the base version
 # number has been incremented)
-__version__ = "0.41.34"
+__version__ = "0.42.0"
 version_info = (
     int(__version__.split(".")[0]),
     int(__version__.split(".")[1]),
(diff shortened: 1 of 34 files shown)


vbrik (approved):
Thanks, Patrick. I appreciate you making this non-trivial change.

PR #282, opened 2024-11-22
update ES script to new 8.16 cluster

alemsh:


resources/progress_to_es.py
```diff
@@ -203,10 +203,22 @@ def main():
...
+                        help='ES oauth2 client id')
+    parser.add_argument('--es_client_secret',default=None,
+                        help='ES oauth2 client secret')
+    parser.add_argument('--token_url',default=None,
+                        help='ES oauth2 realm token url')
     args = parser.parse_args()
 
     logging.basicConfig(level=getattr(logging, args.log_level.upper()), format='%(asctime)s %(levelname)s %(name)s : %(message)s')
 
+    es_api = ClientCredentialsAuth(address='https://elasticsearch.icecube.aq',
```
blinkdog, line 216: Maybe change this to `address=config['ES_ADDRESS']` ?

blinkdog (changes requested):
Looks pretty good! Just one minor suggestion, and it should be ready to merge.

resources/progress_to_es.py
```diff
@@ -203,10 +203,22 @@ def main():
...
+                            token_url=args.token_url,
+                            client_secret=args.es_client_secret,
+                            client_id=args.es_client_id)
```
blinkdog, line 219: One of the CI tests using flake8 is complaining about the under-indent here; line up the others on the same column as address.

blinkdog (changes requested):
Whoops, another small change to pacify CI.

resources/progress_to_es.py
alemsh: Sounds good

resources/progress_to_es.py
alemsh: Ope, good catch updated!

blinkdog (approved):

blinkdog:
LGTM. Please merge when you're ready.

PR #286, opened 2025-03-26
run progress_to_es.py on a bson db backup

dsschult:


resources/progress_to_es.py
```diff
@@ -149,10 +150,15 @@ async def run(self):
...
-            for line in f:
-                catalog_file = json.loads(line)
+        if 'bson' in filename:
+            file_iter = bson.decode_file_iter
+        else:
+            def file_iter(f):
+                for line in f:
+                    yield json.loads(line)
 
+        with gzip.open(filename, 'r') as f:
```
blinkdog, line 160: Looks like it's using `gzip.open()` to open the file regardless; is that intentional?
Like, you can feed it `something.ndjson.gz` or `something.bson.gz`?

blinkdog: Okay, just checking. If you don't need non-gzip file support, it looks fine as-is.

blinkdog (approved):

PR #294, opened 2025-06-20
New project metadata and tooling

blinkdog:
- Added `pyproject.toml` to replace `setup.cfg` and `setup.py`

- TODO: Update for CI/CD workflows


ric-evans:
> Note that I still use a basic `setup.py` file. Not sure if that's actually still required.

It's still required

ric-evans (changes requested):
@blinkdog you'll need to remove	the `__version__ = ...` since the project now builds with scm.

ric-evans (commented):
With these changes, the project is bound for PyPI. Is that what's wanted?

blinkdog:
> With these changes, the project is bound for PyPI. Is that what's wanted?

I'm not sure that any other project would add `lta` as a dependency, so it's probably not that useful to publish to PyPI.

However, publishing to GitHub Container Registry (ghcr.io) is a must; we've got a lot of deployments running in k8s that pull from there. For example:

https://github.com/WIPACrepo/kubernetes/blob/master/rbn-k8s/argocd/long-term-archive/templates/Deployment-pipe0-picker.yaml

That one pulls from `image: ghcr.io/wipacrepo/lta:master`; I'm still looking at incorporating the release actions, so I'm not sure if can keep the `master` tag or if those k8s deployments will need to be updated to the `latest` tag or something.

PR #326, opened 2025-09-22
CI Tests: integration tests

ric-evans:


Dockerfile
```diff
@@ -1,6 +1,8 @@
 # Dockerfile
 FROM almalinux:9
 
+ARG PYTHON=${PYTHON}
```
blinkdog: Does this argument get supplied from the CI builds of Docker images?

And what about from `tools/build-docker`?

blinkdog (changes requested):
Requesting changes to supply PYTHON argument if necessary.

Dockerfile
ric-evans: Looks like you grabbed an old commit, it now has a default.

ric-evans: fyi, the PR is still in draft, so nothing is stable

PR #329, opened 2025-10-01
Replicator Unit Tests [no-bump]

### LINKED ISSUE #309: Unit Tests for Replicator
Currently, https://github.com/WIPACrepo/lta/blob/master/lta/gridftp_replicator.py does not have tests. 

Write tests that will also work after Globus is used.

ric-evans:
closes #309


blinkdog (commented):
LGTM. I like the fixtures, especially the DummyRestClient with the responses; these tests are incredibly clean compared to the old stuff.

Only one nit-pick; I usually like to name the unit test files after the code they are testing (i.e. test_gridftp_replicator.py), but here we're moving away from the gridftp part, so I'm not terribly hung up on the more generic name.

blinkdog (approved):
Meant to approve it!

PR #346, opened 2025-12-18
Update Dockerfile to remove GridFTP dependencies

blinkdog:
@ric-evans sent me a message on Slack about the `Dockerfile` being incompatible with Python 3.14:
>I'm getting a dockerbuild error that I think is coming from a python 3.14 incompatibility
[3:57](https://icecube-spno.slack.com/archives/DNJ89CCJZ/p1766008667858799)
https://github.com/WIPACrepo/lta/actions/runs/20318311549/job/58367818021?pr=345
[3:59](https://icecube-spno.slack.com/archives/DNJ89CCJZ/p1766008758857019)
looking at the dockerfile, there are a few RUNs that reference gridftp
[3:59](https://icecube-spno.slack.com/archives/DNJ89CCJZ/p1766008784171359)
and/or the globus-proxy

I've updated the `Dockerfile` to remove the GridFTP related cruft (certs, tools, etc.), but ironically, it does not fix the Python 3.14 compatibility problem.

The Fedora repo that we add doesn't offer a Python 3.14 package as of this writing.
It does offer 3.13, so I've bumped the default to 3.13.

Removing the cruft is a good idea, so I'm offering the PR anyway. It seems there are three paths forward:

1. Wait. Python 3.14 will show up eventually.
2. Add an Alma9 compatible repo that provides bleeding edge Python releases.
3. Switch to a different container base; maybe `python:3.14-alpine`?

Unless we have an urgent pressing need for Python 3.14 right this second, I favor just waiting.
I think it would be the least work and most secure way to get Python 3.14 compatibility.

さよなら、 GridFTP！ ^^

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,58 +1,14 @@
 # Dockerfile
 FROM almalinux:9
 
-ARG PYTHON=3.12
+ARG PYTHON=3.13
 
-RUN dnf install -y dnf-plugins-core wget && dnf clean all
-
-# install GridFTP tools like globus-url-copy
-RUN wget -O /etc/pki/rpm-gpg/RPM-GPG-KEY-EUGridPMA-RPM-4 \
-    https://dist.eugridpma.info/distribution/igtf/current/GPG-KEY-EUGridPMA-RPM-4 \
-    && rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-EUGridPMA-RPM-4 \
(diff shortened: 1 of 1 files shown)


ric-evans:
Waiting sounds good to me. If there's 3.13, then I assume we can expect 3.14 this year.

ric-evans (approved):

PR #362, opened 2026-01-28
Assign Files to Bundles More Evenly (by size) [minor]

ric-evans:
Previously, we have seen files split into bundles in an unbalanced way. Ex: `115G -> [100G, 15G]`. Now, these will be more evenly distributed while respecting a maximum value. Ex: `115G -> [115G]`.

Refactored: 
- `MAX_BUNDLE_SIZE` has been replaced with `IDEAL_BUNDLE_SIZE`
	- A bundle may be up to 1.2x this size (hardcoded).
- Decrease REST calls to the File Catalog by the number of files (big optimization, like 99%).
	- File sizes are now retrieved when the list of files is asked for, not individually.
- Change file-list pagination (FC queries) to not assume `limit` query-arg is respected.
	- Now, pages are queried until FC returns no more. This requires an additional REST call.
- Lists of files are now passed as dataclass objects (`FileCatalogFile`), rather than dicts.
	- This ensures we know what info is where, and eliminates a couple of intermediate data structures.

### DIFF
--- a/lta/picker.py
+++ b/lta/picker.py
@@ -145,10 +145,10 @@ class Picker(Component):
             )
             return
 
-        # step 3: group those files
+        # step 2: group those files
         packing_spec = self._group_catalog_files_evenly(catalog_files)
 
-        # step 2: bundle those files
+        # step 3: bundle those files
         await self._bundle_files_for_lta(tr, packing_spec, lta_rc)
 
     async def _get_files_page(


blinkdog:
> @blinkdog there are many calls to `'GET' @ f'/api/files?query={query_json}&keys=uuid&limit={self.file_catalog_page_size}&start={page_start}')` (get list of files; paginated) and `GET @ /api/files/{catalog_file_uuid}` (get file size). We can reduce this significantly by requesting each file's size in the first query (`keys=uuid|file_size`). What do you think?

Sure! If we can get the important stuff back in one-shot, I think that's a great optimization.

blinkdog (approved):
LGTM. This makes the bundle sizes nicer and cleans up a lot we-have-types-at-home typing; nice work.

I think this is ready to merge.

PR #368, opened 2026-02-06
Record and Log Stack Traces

### LINKED ISSUE #367: Better Quarantine: record and log stack traces
Nearly all errors (in the form of Python exceptions) in a component lead to quarantine. Recording (in the LTA DB) and logging (stderr) these exceptions is important for debugging and fixing the root causes, but without stack traces, it can be difficult.

At quarantine, stack traces will be:
- Recorded in a  _new_ `reason_details` field in the bundle object.
	- Trimmed to 500 lines, first and last 250.
	- Like `reason`, this is not necessarily limited to quarantine.
	- Upon PATCH, if a `reason == ""`, then override `reason_details` to `""`
- Logged to stderr in its entirety.

Related to #363

ric-evans:
LTA DB:
- adds `reason_details` field
- auto-clears `reason_details` when user clears `reason`

Components:
- log stack trace (in `component.Component.run()`)
	- _Side note:_ I think this was the original intention of the previously written `logger.error(..., exc_info=True)`.

`./ltacmd`
- `--pretty` prints the `reason_details` stacktrace using multiple lines instead of printing `"\n"` chars (this also effectively enables `--json` if it was not included)

closes #367

### DIFF
--- a/lta/lta_cmd.py
+++ b/lta/lta_cmd.py
@@ -312,11 +312,8 @@ def _is_nersc_bundle_record(d: Dict[str, Any]) -> bool:
 async def bundle_ls(args: Namespace) -> ExitCode:
     """List all of the Bundle objects in the LTA DB."""
     response = await args.di["lta_rc"].request("GET", "/Bundles")
-    if args.json or args.extract_print:
-        print_dict_as_pretty_json(
-            response,
-            extract_print_fields=args.extract_print,
-        )
+    if args.json:
+        print_dict_as_pretty_json(response)
     else:
         results = response["results"]
         print(f"total {len(results)}")
@@ -907,8 +904,11 @@ async def request_status(args: Namespace) -> ExitCode:
     response = await args.di["lta_rc"].request("GET", f"/TransferRequests/{args.uuid}")
     res2 = await args.di["lta_rc"].request("GET", f"/Bundles?request={args.uuid}")
     response["bundles"] = await _get_bundles_status(args.di["lta_rc"], res2["results"])
-    if args.json:
-        print_dict_as_pretty_json(response)
+    if args.json or args.extract_print:
+        print_dict_as_pretty_json(
+            response,
+            extract_print_fields=args.extract_print,
+        )
     else:
         # display information about the core fields
         print(f"TransferRequest {args.uuid}")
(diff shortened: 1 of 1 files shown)


blinkdog (changes requested):
A good start!

tests/unit/test_utils.py
```diff
@@ -110,6 +133,87 @@ async def test_110_quarantine_exc_reason() -> None:
...
+                f"{'    ~~~~~~~^^'+nl if sys.version_info >= (3, 13) else ''}"
+                f'  File "{__file__}", '
+                "line 162, in my_func\n"
+                "    _inner_func()\n"
+                f"{'    ~~~~~~~~~~~^^'+nl if sys.version_info >= (3, 13) else ''}"
+                f'  File "{__file__}", '
+                "line 159, in _inner_func\n"
+                '    raise ValueError("nope")\n'
+                "ValueError: nope\n"
+            ),
```
blinkdog: Maybe refactor this into a triple-quote string at the end of the test source file?
Here, it'd be:

```python
traceback = TRACEBACK_PY_OLD
if sys.version_info >= (3, 13):
    traceback = TRACEBACK_PY_313

"reason_details": traceback,
```

Or maybe:

```python
"reason_details": _choose_traceback(sys.version_info),
```

And at the end of the source file:

```python
TRACEBACK_PY_OLD = """Traceback (most recent call last):
... stuff omitted here ...
ValueError: nope
"""

TRACEBACK_PY313 = """Traceback (most recent call last):
... different stuff omitted here ...
ValueError: nope
"""
```

It's a little more verbose, but it may read a little more clearly to future-us 😅

blinkdog (changes requested):
Ack, I didn't submit it correctly the first time... ごめん

tests/unit/test_utils.py
ric-evans: I gotchu. That's a good point

lta/lta_cmd.py
```diff
@@ -107,9 +112,50 @@ def print_catalog_record_as_line(d: Dict[str, Any]) -> None:
...
+    """
+
+    def _print_dict(_out: Dict[str, Any]):
+        print(json.dumps(_out, indent=4, sort_keys=True))
+
+    if not pretty_multilines:
+        _print_dict(d)
+        return
+
+    # now, let's see if there are any multiline fields to print separately...
```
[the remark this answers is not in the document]
ric-evans: updated with [`25e7a0f` (this PR)](https://github.com/WIPACrepo/lta/pull/368/commits/25e7a0f21db92731e3b73db838e4c429656d25bc)

PR #370, opened 2026-02-13
Prometheus Metrics: Components

### LINKED ISSUE #350: Prometheus Monitoring: use `wipac_dev_tools.prometheus_tools`
This was first mentioned in #349.

This package is implemented in iceprod (see https://github.com/WIPACrepo/iceprod/blob/ee670a0603352d54764e1c92d8bc5afe7bd67cdf/iceprod/server/grid.py#L21), so we know it works in production.

ric-evans:
Prometheus metrics are now (re-)implemented for the components. I drew inspiration from how @dsschult used Prometheus for iceprod in https://github.com/WIPACrepo/iceprod/blob/ee670a0603352d54764e1c92d8bc5afe7bd67cdf/iceprod/server/grid.py#L21.

**Metrics:**
- Histogram for per-bundle (or per-transfer-request) work latency
	- Only successful "work instances" are recorded
- Counters for per-bundle (or per-transfer-request) successes and failures, each

**Major change:**
The main work loop is now centralized in `Component._do_work()`, which calls the child class `_do_work_claim()`. This allows us to avoid duplicating the boilerplate Prometheus setup. This is important so we can compare component metrics without future code drift.

**Other changes:**
- The transfer-request components (picker and locator) now use the common `quarantine_now()`.
	- The REST server has been updated to do the same auto-clear for the `'reason_details'` field when the `'reason'` field is reset, as is done for bundles.
- Removed the unused `last_work_end_timestamp` and `last_work_begin_timestamp`
- `quarantine_now()` now only takes exceptions (not sure why I didn't do this when I made it; it simplifies tests.)
- Replaced the deprecated `datetime.utcnow()` usage.

closes #350

### DIFF
--- a/lta/component.py
+++ b/lta/component.py
@@ -213,7 +213,7 @@ class Component:
         prometheus_histogram = self.prometheus.histogram(
             "lta_single_work_latency_seconds",
             "LTA component: time taken to process a single work item (only successes are recorded)",
-            buckets=HistogramBuckets.TENMINUTE,  # TODO: do we want to make this configurable?
+            buckets=HistogramBuckets.HOUR,  # common across all components
         )
 
         for i in itertools.count():


lta/component.py
```diff
@@ -40,16 +42,48 @@
...
+        self._done = True
+        self._success_counter.inc()
+        self._histogram.observe(time.monotonic() - self._start_ts)
+
+    def record_failure(self):
+        """Record a failed work -- this should only be called at the end."""
+        if self._done:
+            raise RuntimeError("Cannot record result twice.")
+        self._done = True
+
```
blinkdog: Should there be a `self.failure_counter.inc()` here?
Do failures also show up in the `self._histogram`, or is that just successes?

lta/desy_mirror_replicator.py
```diff
@@ -9,12 +9,12 @@
...
 
-from prometheus_client import Counter, Gauge, start_http_server
+from prometheus_client import start_http_server
 from rest_tools.client import RestClient
 from wipac_dev_tools import strtobool
 
-from .utils import quarantine_bundle
-from .component import COMMON_CONFIG, Component, now, work_loop
+from .component import COMMON_CONFIG, Component, work_loop, PrometheusResultTracker
+from .utils import now, quarantine_now
```
blinkdog: Move `.utils` to maintain alphabetical order.

lta/lta_cmd.py
```diff
@@ -24,7 +24,7 @@
 import urllib
 from wipac_dev_tools import from_environment
 
-from lta.component import now
+from lta.utils import now
```
blinkdog: Move `lta.utils` to maintain alphabetical order.

lta/transfer_request_finisher.py
```diff
@@ -71,22 +67,12 @@ def _expected_config(self) -> Dict[str, Optional[str]]:
...
-                sys.exit()
-        load_gauge.labels(component='transfer_request_finisher', level='bundle', type='work').set(load_level)
-        self.logger.info("Ending work on Bundles.")
-
-    async def _do_work_claim(self, lta_rc: RestClient) -> bool:
-        """Claim a bundle and perform work on it."""
+    async def _do_work_claim(
+        self,
+        lta_rc: RestClient,
+        prom_tracker: PrometheusResultTracker,
```
blinkdog: Did we actually record a result here?

lta/utils.py
```diff
@@ -1,17 +1,42 @@
...
+from typing import Literal
 
 from rest_tools.client import RestClient
 
-from lta.component import now
-from lta.lta_types import BundleType
+from lta.lta_types import BundleType, TransferRequestType
+
+
+LtaObjectType = Literal["BUNDLE", "TRANSFER_REQUEST"]
```
blinkdog: Replace with constants; see comment on `quarantine_now` function.

blinkdog (changes requested):
A bunch of nitpicks about imports (ごめん; sorry), but also a few things that are fix (good catches) or push-back (bad catches, because I didn't understand it correctly when I reviewed).

Very good first pass!

dsschult (commented):
General comment on metric names.  You may want to prefix metric names with `lta_` if it's not a common metric with other programs (like `http_request_duration_seconds`).

lta/component.py
```diff
@@ -156,8 +195,61 @@ def _expected_config(self) -> Dict[str, Optional[str]]:
...
+            self.logger.info(f"Requesting work on #{i} (0-indexed)...")
+            prom_tracker = PrometheusResultTracker(
+                prom_counter.labels({"work": "success"}),
+                prom_counter.labels({"work": "failure"}),
+                prometheus_histogram,
+                time.monotonic(),  # instantiate now for accurate timestamp
+            )
+
+            ret = await self._do_work_claim(lta_rc, prom_tracker)
+
```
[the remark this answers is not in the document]
blinkdog: Originally, I thought it was more a logic error than a monitoring error; like the component found its way through a code path that didn't trigger success or failure.

OTOH, we might have components that babysit 'Is it done yet?' processes asynchronously. In that case, 'kick it back and check on it later' isn't really success (it didn't finish and move on to the next stage), nor is it failure (it didn't fail, it just isn't finished yet).

Do we need a third counter type? Indeterminate? In-Progress? Incomplete?

Or do we just say tough cookies, sometimes you just don't get a success/failure count.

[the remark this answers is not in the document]
blinkdog: There are two cases I think of off the top of my head:

1. NerscMover could return early if it detects HPSS is down. In that case it doesn't actually pull work from the LTA DB. `_do_work_claim()` is called, but this isn't really success or failure, just can't-work-right-now.

https://github.com/WIPACrepo/lta/pull/370/changes#diff-1d6e6e5c213251ec5f54a5f0640f8c541279bd7d598f87d95bb2cac41fe4a6b6L107

2. Today's GlobusReplicator sits on top of a transfer until it is complete

But one might imagine a different construction that pulled a bundle, checked the Globus API, found the transfer in progress, 'threw the bundle back' for checking later (no status change). In that case, the transfer hasn't succeeded (isn't complete) and hasn't failed (is still transferring).

This doesn't apply to GlobusReplicator today, it either succeeds or fails as the transfer does.

lta/transfer_request_finisher.py
ric-evans: No, no we did not :rofl: 

got it

lta/component.py
```diff
@@ -156,8 +195,61 @@ def _expected_config(self) -> Dict[str, Optional[str]]:
...
+        "finished work counts by success/failure",
+        labels=["work"],
+        finalize=False,
+    ))
+    async def _do_work(self, prom_counter: Counter, lta_rc: RestClient) -> None:
+        """Perform a work cycle for this component."""
+        prometheus_histogram = self.prometheus.histogram(
+            "single_work_latency_seconds",
+            "time taken to process a single work item (only successes are recorded)",
+            buckets=HistogramBuckets.TENMINUTE,  # TODO: do we want to make this configurable?
```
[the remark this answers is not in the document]
blinkdog: Ahhh, okay, I see what you mean now. Yeah, there will be components with a slightly different time scales.

Components that move a lot of data (Bundler, GlobusReplicator, NerscMover), those that check a lot of data (SiteMoveVerifier, NerscVerifier), and those that push around some metadata (RateLimiter, Deleter).

Is this collecting work in aggregate across all components? Or are these metrics captured per-component?

I mean, are they sharing the same buckets, so we pretty much have to keep them the same because they'll be living together on the same graph?

Or is there a graph per component, so we need the correct size bucket for each graph?

PR #373, opened 2026-02-20
Prometheus Metrics: REST Server Statuses

### LINKED ISSUE #371: Prometheus Metrics: Status Gauge on REST Server
As @blinkdog mentioned, it would be useful to have Prometheus gauges for the LTA bundle statuses. This would be best done by querying the LTA database. I picture dedicated endpoint(s) on the REST server. More research is needed.

ric-evans:
New Prometheus Metrics:
- Gauge: `lta_status`
	- A real-time reflection of the LTA database (runs in a background task)
- Counter: `lta_bundle_claims_total`
	- An only-increasing count of all claims made on bundles
- Counter: `lta_status_writes_total`
	- An only-increasing count of all writes of a status
	- Note: if, for some reason, a status-write writes the same status as before, the counter still goes up—but this should not be a problem, given the components are the main "user".

closes #371

### DIFF
--- a/lta/rest_server.py
+++ b/lta/rest_server.py
@@ -849,7 +849,7 @@ def start(
     return server
 
 
-def _log_if_time(logging_timer: IntervalTimer, logger: logging.Logger) -> Callable:
+def _logger_if_time(logging_timer: IntervalTimer, logger: logging.Logger) -> Callable:
     # Note - we could go more generic (pass logger.info, etc.), but this is a simple case
     if logging_timer.has_interval_elapsed():
         return logger.info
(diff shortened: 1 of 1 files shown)


lta/rest_server.py
```diff
@@ -108,15 +114,34 @@ def unique_id() -> str:
...
-# make module-level so same histogram is shared within this process (else, overwrites)
 PROMETHEUS_HISTOGRAM = prometheus_client.Histogram(
     'http_request_duration_seconds',
     'HTTP request duration in seconds',
     labelnames=('method', 'route', 'status'),
     buckets=prometheus_tools.HistogramBuckets.HTTP_API,
 )
 
+PROMETHEUS_STATUS_GAUGE = prometheus_client.Gauge(
+    "lta_status_count",
```
dsschult: I recommend not using `_count`, as that has a special meaning in prometheus.  Especially as this is a Gauge, not a Counter.

lta/rest_server.py
```diff
@@ -786,10 +847,65 @@ def start(debug: bool = False) -> RestServer:
     return server
 
 
+async def status_poller(
```
dsschult: Do we want this to run directly on the server as a side task, or to run as a separate cron job in k8s?

dsschult (approved):

